PR #124, opened 2018-08-18
Copy LE certs for ZNC

d2dyno:


liaralabs:
I'd be happy to include this, but I'd like to see a similar hook in the letsencrypt script in the event that letsencrypt hasn't been setup yet but ZNC is already installed.

d2dyno:
@liaralabs I have made the requested changes to the LE install file.

Issue #138, opened 2018-09-07
Change plexpy to Tautulli


ooohhh:
Plexpy no longer exists and has been replaced by Tautulli
https://github.com/Tautulli/Tautulli

PR #140, opened 2018-09-08
Update Plexpy to Tautulli

ooohhh:
Resovles #138


ooohhh:
Panel update: https://github.com/liaralabs/quickbox_dashboard/pull/5

asafsarid:
Does this support a case where someone already had plexpy installed?
I think you might need to add logic to remove/stop plexpy before installing Tautulli. Or maybe adding it to the update/upgrade script.

ooohhh:
It doesn't, I missed the update/upgrade scripts. I'll add them to the PR.

Thanks for noticing it.

liaralabs:
Is there a reason you went with release vs. github? I haven't tested anything other than the github version, but the general reason that there isn't an upgrade script for Tautulli is because the application keeps itself up to date (I don't rewrite things that already have methods). I don't know if the release versions are capable of keeping themselves up to date?

I haven't had time to test the update path yet, but it should be okay -- only thing I notice is that a permanent redirect from /plexpy -> /tautulli would be good to ease people into the transition.

ooohhh:
For stability I guess, the master on github is not always stable. I think all apps should be installed via releases and not the latest code on github. If you disagree I can change it, or maybe we can add an option for the user to decide.

Regarding the upgrade script, Tautulli handles it's upgrades fine, especially on releases. it was mainly for the issue raised by @asafsarid for users that are running plexpy. But since I saw an upgrade script for Ombi & Jackett which should also upgrade themselves I thought I should add one too.

Regarding the PR, I noticed a few things that I need to fix in the config.ini when moving from plexpy to tautulli, I'll get to it today.

liaralabs:
Sounds good. I'll look for your next commits. I'm fine to stick with the release version as long as the updates continue to work without the upgrade script.

I appreciate you taking the time to fix this, the reason I haven't updated is mostly because it's cosmetic (and because I can barely spell tautulli)

ooohhh:
The last commits change the following:
1. Remove the upgrade script and leave the update script only when upgrading from plexpy to tautulli
2. Remove basic auth for tautulli as it has it's own authentication and the basic auth blocks the android app.
3. Added a redirect from /plexpy to /tautulli

PR #155, opened 2018-10-20
Update vsftpd with LE certs when available

d2dyno:
Also verify only one ssl cert host is in the cert directory.


liaralabs:
Can you rebase against the changes I just pushed to combat multiple LE certs existing in the same environment? My fix will just use the default host to apply certs, rather than skipping the config -- should be a bit more resilient.

d2dyno:
@liaralabs please review the latest commits, across all files that used this method.

PR #158, opened 2018-10-26
Massively simplify rclone installation

d2dyno:
Use one-liner from https://rclone.org/downloads/; use https now instead of http link before; do NOT make system service, unless rclone install proceeds without issue.


liaralabs:
Looks good, however the cat command into the systemd service has bad syntax. I generally avoid formatting cat commands nicely to fit within the script as the cat command will read each line literally. Further, the EOF needs to be on the last line *without* any spaces/tabs in front.

As you can see from the github preview, the syntax highlighting breaks after the cat command.

d2dyno:
I missed that syntax error, my apologies. Please review the updated file.

PR #173, opened 2018-12-16
Add cronjob renewal to LE

d2dyno:
I believe this is all that is required. If we installed via `install`, cronjob would be added automatically, but since we use `install-cert` it is not.


liaralabs:
Sorry for leaving this open for a year -- this isn't necessary because acme installs the cronjob during the initial install (when we grab acme and bash the install script). You can verify that the cronjob has been added with `crontab -l` as root.

PR #254, opened 2019-10-27
Jellyfin support, closes #220

### LINKED ISSUE #220: Add Jellyfin Support
[Jellyfin Repo](https://github.com/jellyfin/jellyfin)

Please consider Jellyfin as one of the apps that can be installed with swizzin. It's a port from Emby 3.5.2 and is pretty stable on my testing (performs better than Plex in terms of direct Play).

Thanks alot.

SavageCore:
Based on the work of @userdocs


scripts/install/jellyfin.sh
```diff
@@ -0,0 +1,298 @@
...
+  <SendOnlyMatchedHost>true</SendOnlyMatchedHost>
+  <ClientDiscoveryIntervalSeconds>60</ClientDiscoveryIntervalSeconds>
+  <BlastAliveMessageIntervalSeconds>1800</BlastAliveMessageIntervalSeconds>
+</DlnaOptions>
+CONFIG
+fi
+
+# Configure the nginx proxypass using positional parameters.
+if [[ -f /install/.nginx.lock ]]; then
+  bash "/usr/local/bin/swizzin/nginx/jellyfin.sh" "${app_port_http}" "${app_port_https}"
```
liaralabs: nginx scripts should not be designed to be need arguments to build the correct configuration. This is important for the `box upgrade nginx` script (which also doubles as an nginx conf repair) as the final step of the repair process is to re-run all nginx scripts for installed apps, and it does so without arguments.

Thus, the nginx script should be crafted in a way that it detects the info it needs and pulls it automatically

scripts/nginx/jellyfin.sh
```diff
@@ -0,0 +1,54 @@
...
+# Licensed under GNU General Public License v3.0 GPL-3 (in short)
+#
+#   You may copy, distribute and modify the software as long as you track
+#   changes/dates in source files. Any modifications to our software
+#   including (via compiler) GPL-licensed code must also be made available
+#   under the GPL along with build & install instructions.
+# username="$(cat /root/.master.info | cut -d: -f1)"
+username="$(cut </root/.master.info -d: -f1)"
+ip_address="$(curl -s4 icanhazip.com)"
+#
```
liaralabs: Assign `app_port_http` and `app_port_https` by grepping our system.xml here:

```
app_port_http=$(grep -oP "<PublicPort>\K[^<]+" /home/${username}/.config/Jellyfin/config/system.xml)
app_port_https=$(grep -oP "<PublicHttpsPort>\K[^<]+" /home/${username}/.config/Jellyfin/config/system.xml)
```

scripts/nginx/jellyfin.sh
```diff
@@ -0,0 +1,54 @@
...
+  sed -r 's#<string>0.0.0.0</string>#<string>127.0.0.1</string>#g' -i "/home/${username}/.config/Jellyfin/config/system.xml"
+fi
+#
+if [[ -z "$1" && -f /install/.jellyfin.lock ]]; then
+  service jellyfin stop
+  app_port_https="$(sed -rn 's#(.*)<HttpsPortNumber>(.*)</HttpsPortNumber>#\2#p' "/home/${username}/.config/Jellyfin/config/system.xml")"
+  sed -r 's#<WanDdns>(.*)</WanDdns>#<WanDdns>https://'"${ip_address}"'/jellyfin</WanDdns>#g' -i "/home/${username}/.config/Jellyfin/config/system.xml"
+  sed -r 's#<string>0.0.0.0</string>#<string>127.0.0.1</string>#g' -i "/home/${username}/.config/Jellyfin/config/system.xml"
+  service jellyfin start
+fi
```
liaralabs: I don't see the need for an `-n` and a `-z` let's just assume `-z` always because I currently see no use case for calling the nginx script with an argument when the standard for all other nginx scripts is without.

scripts/install/jellyfin.sh
userdocs: I am pretty sure the file this was based on already dealt with this:

https://github.com/userdocs/swizzin/blob/master/scripts/nginx/jellyfin.sh

userdocs:
Having relooked at my original code and some of this pull request I think there is more testing that needs to be done before implementing. Some code has been pulled across that I would consider obsolete. If you are happy to do that I will happily assist to get jellydin pulled.

SavageCore:
> 
> 
> Having relooked at my original code and some of this pull request I think there is more testing that needs to be done before implementing. Some code has been pulled across that I would consider obsolete. If you are happy to do that I will happily assist to get jellyfin pulled.

I welcome your input! What were you referring to [here](https://github.com/liaralabs/swizzin/issues/220#issuecomment-528059412)? I've been using it and the URLs seem fine.

userdocs:
To start, i updated my code for testing. Generically it works but there seems to be new base url option making previous settings obsolete. I only used to the wandns setting to overcome an ssl issue I was experiencing. It seems no longer valid as a setting or relevant to the functionality of the program.

userdocs:
We also potentially have issues using the official jellyfin app with this forced https setup. Please see 
https://github.com/jellyfin/jellyfin-android/issues/193 and https://github.com/jellyfin/jellyfin-android/issues/217

I cannot get the android jellyfin app to connect using a self signed https cert

SavageCore:
> To start, i updated my code for testing. Generically it works but there seems to be new base url option making previous settings obsolete. I only used to the wandns setting to overcome an ssl issue I was experiencing. It seems no longer valid as a setting or relevant to the functionality of the program.

Could be, not tested thoroughly, just enough to have it working for myself.

> We also potentially have issues using the official jellyfin app with this forced https setup. Please see
> [jellyfin/jellyfin-android#193](https://github.com/jellyfin/jellyfin-android/issues/193) and [jellyfin/jellyfin-android#217](https://github.com/jellyfin/jellyfin-android/issues/217)
> 
> I cannot get the android jellyfin app to connect using a self signed https cert

I wonder if this is really a blocking issue? How many run self signed with letsencrypt built in...?

I've invited you to my fork as a contributor if you'd like to contribute to this branch. I think that is the right way?

userdocs:
Jellyfin 10.4.1 support https://github.com/userdocs/swizzin

userdocs:
Jellyfin 10.4.2 support https://github.com/userdocs/swizzin/commit/05f5f7a845e4811cf7f194b9668995c384a87c12

NathanBlais:
When are we getting this added to the main branch?

PR #280, opened 2020-04-11
new application: LibreSpeed

hwcltjn:
This commit adds 3 new files for the purpose of including the lightweight [LibreSpeed](https://github.com/librespeed/speedtest/) speed test to Swizzin

LibreSpeed is installed to `/srv/librespeed`

No database or configuration required.

### DIFF
--- a/scripts/install/librespeed.sh
+++ b/scripts/install/librespeed.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+# Librespeed installer for swizzin
+# Author: hwcltjn
+
+if [[ ! -f /install/.nginx.lock ]]; then
+  echo "ERROR: Web server not detected. Please install nginx and restart panel install."
+  exit 1
+fi
+
+if [[ -f /tmp/.install.lock ]]; then
+  OUTTO="/root/logs/install.log"
+else
+  OUTTO="/root/logs/swizzin.log"
+fi
+
+lspdpath='/srv/librespeed'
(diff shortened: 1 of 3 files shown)


scripts/install/librespeed.sh
```diff
@@ -0,0 +1,61 @@
...
+
+lspdpath='/srv/librespeed'
+htuser='www-data'
+htgroup='www-data'
+
+function _installLibreSpeed1() {
+	mkdir $lspdpath
+	git clone https://github.com/librespeed/speedtest.git $lspdpath >/dev/null 2>&1
+	cp $lspdpath/example-singleServer-gauges.html $lspdpath/index.html
+	swizname=$(sed -ne '/server_name/{s/.*server_name //; s/[; ].*//; p; q}' /etc/nginx/sites-enabled/default)
```
liaralabs, line 24: By default this is going to resolve to `_` unless a user has installed Let's Encrypt or manually edited the config file on their own. Can we have some sort of fallback in the event that `swizname=_` rather than in the event that the variable is empty? The `server_name` variable is not likely to be empty.

hwcltjn: How about `[ ! -z $swizname ] && [ $swizname != "_" ]` ?

liaralabs: Yeah, I believe that would just about solve it.

PR #285, opened 2020-04-23
PHP 7.4+ Support and 20.04 testing

flying-sausages:
this PR attempts at resolving the issue with PHP version guessing by abstracting the repetitive logic into one central place, and working with parsing rather than checking against hardcoded files.

I also plan to test this on 20.04, and introduce any changes that are necessary to have it run there.

Please note that I am ignorant to the history of the project as I just found it a couple days ago, and am very happy to learn of some of the possibly historical or back-compat choices made in the code.

### DIFF
--- a/scripts/install/nginx.sh
+++ b/scripts/install/nginx.sh
@@ -84,7 +84,7 @@ for version in $phpv; do
   phpenmod -v $version opcache
 done
 
-sock=$(ls /lib/systemd/system | grep php | sort | tail -1 | grep -o -P '.*(?=.service)')
+sock=$(ls /lib/systemd/system | grep php | grep fpm | sort | tail -1 | grep -o -P '.*(?=.service)')
 
 rm -rf /etc/nginx/sites-enabled/default
 cat > /etc/nginx/sites-enabled/default <<NGC
--- a/scripts/install/nginx.sh
+++ b/scripts/install/nginx.sh
@@ -57,7 +57,7 @@ else
   geoip=php-geoip
 fi
 
-if [[ $codename =~ ("bionic"|"buster") ]]; then
+if [[ $codename =~ ("bionic"|"buster"|"focal") ]]; then
   mcrypt=
 else
   mcrypt=php-mcrypt
@@ -84,7 +84,11 @@ for version in $phpv; do
   phpenmod -v $version opcache
 done
 
-sock=$(ls /lib/systemd/system | grep php | grep fpm | sort | tail -1 | grep -o -P '.*(?=.service)')
+
+. /etc/swizzin/sources/functions/php
+phpversion=$(php_service_version)
+sock="php${phpversion}-fpm"
+echo "Using ${sock} for the service"
 
 rm -rf /etc/nginx/sites-enabled/default
 cat > /etc/nginx/sites-enabled/default <<NGC
--- a/scripts/install/nginx.sh
+++ b/scripts/install/nginx.sh
@@ -84,15 +84,7 @@ for version in $phpv; do
   phpenmod -v $version opcache
 done
 
-if [[ -f /lib/systemd/system/php7.3-fpm.service ]]; then
-  sock=php7.3-fpm
-elif [[ -f /lib/systemd/system/php7.2-fpm.service ]]; then
-  sock=php7.2-fpm
-elif [[ -f /lib/systemd/system/php7.1-fpm.service ]]; then
-  sock=php7.1-fpm
-else
-  sock=php7.0-fpm
-fi
+sock=$(ls /lib/systemd/system | grep php | sort | tail -1 | grep -o -P '.*(?=.service)')
 
 rm -rf /etc/nginx/sites-enabled/default
 cat > /etc/nginx/sites-enabled/default <<NGC
--- a/scripts/install/nginx.sh
+++ b/scripts/install/nginx.sh
@@ -88,7 +88,7 @@ done
 . /etc/swizzin/sources/functions/php
 phpversion=$(php_service_version)
 sock="php${phpversion}-fpm"
-echo "Using ${sock} for the service"
+echo "Using ${sock} in the nginx config"
 
 rm -rf /etc/nginx/sites-enabled/default
 cat > /etc/nginx/sites-enabled/default <<NGC
--- a/scripts/update/nginx.sh
+++ b/scripts/update/nginx.sh
@@ -42,15 +42,8 @@ if [[ $phpv =~ "7.1" ]]; then
   fi
 fi
 
-if [[ -f /lib/systemd/system/php7.3-fpm.service ]]; then
-  sock=php7.3-fpm
-elif [[ -f /lib/systemd/system/php7.2-fpm.service ]]; then
-  sock=php7.2-fpm
-elif [[ -f /lib/systemd/system/php7.1-fpm.service ]]; then
-  sock=php7.1-fpm
-else
-  sock=php7.0-fpm
-fi
+phpversion=$(php_service_version)
+sock="php${phpversion}-fpm"
 
 for version in $phpv; do
   if [[ -f /etc/php/$version/fpm/php.ini ]]; then
@@ -67,6 +60,7 @@ for version in $phpv; do
   fi
 done
 
+##TODO
 if [[ -f /lib/systemd/system/php7.3-fpm.service ]]; then
   v=$(find /etc/nginx -type f -exec grep -l "fastcgi_pass unix:/run/php/php7.3-fpm.sock" {} \;)
   if [[ -z $v ]]; then
--- a/scripts/update/nginx.sh
+++ b/scripts/update/nginx.sh
@@ -42,6 +42,7 @@ if [[ $phpv =~ "7.1" ]]; then
   fi
 fi
 
+. /etc/swizzin/sources/functions/php
 phpversion=$(php_service_version)
 sock="php${phpversion}-fpm"
 
@@ -60,7 +61,6 @@ for version in $phpv; do
   fi
 done
 
-. /etc/swizzin/sources/functions/php
 phpversion=$(php_service_version)
 
 v=$(find /etc/nginx -type f -exec grep -l "fastcgi_pass unix:/run/php/php${phpversion}-fpm.sock" {} \;)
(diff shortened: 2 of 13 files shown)


flying-sausages:
I have now been able to run setup.sh, install nginx, rtorrent and rutorrent with no issues.

flying-sausages:
panel installs just fine

flying-sausages:
deluge from repo seems to install fine but the python scripts need tro be called from python3 on focal and I'm not sure they're compatible.

setup.sh
```diff
@@ -53,7 +53,7 @@ _os() {
 function _preparation() {
   echo "Updating system and grabbing core dependencies."
   if [[ $distribution = "Ubuntu" ]]; then
-    echo "Checking enabled repos"
+    echo "Fetching updates from enabled repos"
```
liaralabs, line 56: Please don't change this. This is literally a check for enabled repos and to enable universe/multiverse if they are not already enabled

setup.sh
```diff
@@ -75,7 +75,7 @@ function _preparation() {
   apt-get -y install whiptail git sudo curl wget lsof fail2ban apache2-utils vnstat tcl tcl-dev build-essential dirmngr apt-transport-https ${pipv} >> ${log} 2>&1
   nofile=$(grep "DefaultLimitNOFILE=500000" /etc/systemd/system.conf)
   if [[ ! "$nofile" ]]; then echo "DefaultLimitNOFILE=500000" >> /etc/systemd/system.conf; fi
-  echo "Cloning swizzin repo to localhost"
+  echo "Installing swizzin files from git"
```
liaralabs, line 78: I'm also not really interested in changing this

setup.sh
```diff
@@ -87,7 +87,7 @@ function _nukeovh() {
   if [[ -n $grsec ]]; then
     echo
     echo -e "Your server is currently running with kernel version: $(uname -r)"
-    echo -e "While not it is not required to switch, kernels with grsec are not recommend due to conflicts in the panel and other packages."
+    echo -e "While it is not required to switch, kernels with grsec are not recommend due to conflicts in the panel and other packages."
```
liaralabs, line 90: This is a typo, and I would allow this

setup.sh
```diff
@@ -172,7 +172,14 @@ function _adduser() {
...
   if grep ${user} /etc/sudoers.d/swizzin >/dev/null 2>&1 ; then echo "No sudoers modification made ... " ; else	echo "${user}	ALL=(ALL:ALL) ALL" >> /etc/sudoers.d/swizzin ; fi
-  echo "D /var/run/${user} 0750 ${user} ${user} -" >> /etc/tmpfiles.d/${user}.conf
+  
+  if [[ $codename =~ "focal" ]] ; then
+    echo "D /run/${user} 0750 ${user} ${user} -" >> /etc/tmpfiles.d/${user}.conf
+  else
+    echo "D /var/run/${user} 0750 ${user} ${user} -" >> /etc/tmpfiles.d/${user}.conf
+
+  fi
+
```
liaralabs, line 182: I believe this can just be applied to all supported distros and it's not needed to only do for focal. That said, this change needs to be reflected in the rtorrent script for the unix socket.

setup.sh
```diff
@@ -33,6 +33,8 @@ _os() {
   if [ ! -d /install ]; then mkdir /install ; fi
   if [ ! -d /root/logs ]; then mkdir /root/logs ; fi
   export log=/root/logs/install.log
+  echo "To follow the progress of the script alongside, run \`sudo watch -n1 tail ${log}\` in another terminal session."
```
liaralabs, line 36: You don't need to `watch` a tail. You can use `tail -f`

setup.sh
```diff
@@ -61,11 +63,21 @@ function _preparation() {
   fi
   apt-get -q -y update >> ${log} 2>&1
   apt-get -q -y upgrade >> ${log} 2>&1
-  apt-get -q -y install whiptail git sudo curl wget lsof fail2ban apache2-utils vnstat tcl tcl-dev build-essential dirmngr apt-transport-https python-pip >> ${log} 2>&1
+
+  if [[ $codename =~ ("focal") ]]; then 
+      pipv='python3-pip'
+      else
+      pipv='python-pip'
+  fi
```
liaralabs, line 71: pip is not a direct dependency of swizzin, thus it does not need to be installed here.

scripts/update/nginx.sh
```diff
@@ -42,15 +42,8 @@ if [[ $phpv =~ "7.1" ]]; then
...
-if [[ -f /lib/systemd/system/php7.3-fpm.service ]]; then
-  sock=php7.3-fpm
-elif [[ -f /lib/systemd/system/php7.2-fpm.service ]]; then
-  sock=php7.2-fpm
-elif [[ -f /lib/systemd/system/php7.1-fpm.service ]]; then
-  sock=php7.1-fpm
-else
-  sock=php7.0-fpm
-fi
+phpversion=$(php_service_version)
```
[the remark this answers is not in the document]
flying-sausages: Done, how do I mark as resolved from this view ._.

flying-sausages:
still draft, just doesn't need it in the name as it's clear

sources/functions/rtorrent
```diff
@@ -6,23 +6,25 @@ export codename=$(lsb_release -cs)
 
 function whiptail_rtorrent () {
   if [[ -z $rtorrentver ]] && [[ -z $1 ]]; then
+    repoversion=$(apt-cache policy rtorrent | grep focal -B 1 | head -1 | cut -c 6- | cut -c -5)
     if [[ ! ${codename} =~ ("xenial"|"jessie") ]]; then
-      function=$(whiptail --title "Install Software" --menu "Choose an rTorrent version:" --ok-button "Continue" --nocancel 12 50 3 \
+      function=$(whiptail --title "Install Software" --menu "Choose an rTorrent version:" --ok-button "Continue" --nocancel 20 50 5 \
+                  repo "(${repoversion})"\
```
flying-sausages, line 12: The formatting of this should be checked out, I'm not sure I like the way it looks but functionally it does the job
image

setup.sh
flying-sausages: It seems that this just runs fine with this change on my system, without a modification on the rtorrent side

flying-sausages:
rtorrent from repo and rutorrent seem to install just fine, my environment had issues with building libtorrent from source, please double check that

- [ ] Check libtorrent builds from source fine

sources/functions/rtorrent
flying-sausages: Decreased height 
image

scripts/update/nginx.sh
```diff
@@ -67,30 +60,15 @@ for version in $phpv; do
...
-  fi
+. /etc/swizzin/sources/functions/php
+phpversion=$(php_service_version)
+
+v=$(find /etc/nginx -type f -exec grep -l "fastcgi_pass unix:/run/php/php${phpversion}-fpm.sock" {} \;)
+if [[ -z $v ]]; then
+  oldv=$(find /etc/nginx -type f -exec grep -l "fastcgi_pass unix:/run/php/" {} \;)
+  for upgrade in $oldv; do
+    sed -i "s/fastcgi_pass .*/fastcgi_pass unix:\/run\/php\/php${phpversion}-fpm.sock;/g" $upgrade
+  done
```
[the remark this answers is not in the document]
flying-sausages: - [ ] Check the replacement works as expected

sources/functions/rtorrent
flying-sausages: Code removed from PR for overreach

scripts/install/nginx.sh
```diff
@@ -57,7 +57,7 @@ else
   geoip=php-geoip
 fi
 
-if [[ $codename =~ ("bionic"|"buster") ]]; then
+if [[ $codename =~ ("bionic"|"buster"|"focal") ]]; then
```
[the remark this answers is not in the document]
liaralabs: I think you are actually correct here. If mcrypt is gone then the list of distros which support it will soon diminish to none once Jessie and Xenial are EOL

setup.sh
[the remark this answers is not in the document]
liaralabs: You're probably right here. I can't remember why I added pip as a swizzin depend -- it is a bit out of place here

setup.sh
```diff
@@ -61,11 +63,21 @@ function _preparation() {
...
+
+  echo "Installing dependencies"
+  # this apt-get should be checked and handled if fails, otherwise the install borks.
```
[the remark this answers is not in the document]
liaralabs: This command *should* be bulletproof, but there are times when this command will fail due to apt locks as I mentioned above and it would be best to avoid this scenario as the installer will continue along blissfully unaware that needed packages are missing.

setup.sh
liaralabs: When /var/run is deprecated this will cause a failure, *eventually*

sources/functions/php
```diff
@@ -1,37 +1,29 @@
...
-  fi
-}
\ No newline at end of file
+    version=$(php_v_from_nginxconf)
+    serv="php${version}-fpm.service"
+    #serv=(ls /etc/systemd/system | grep php7 | tail -1)
+    systemctl restart $serv
+}
+
+# function disable_other_phps () {
```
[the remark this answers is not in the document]
liaralabs: Yeah, I can work on this as part of my contributions to the update

sources/functions/php
```diff
@@ -1,37 +1,29 @@
...
+#Returns for example '7.4'
+function php_v_from_nginxconf () { 
+    nginxconfpath='/etc/nginx/sites-enabled/default'
+    version=$(cat $nginxconfpath | grep -o -P '(?<=/php/php).*(?=-fpm.sock)')
+    echo "${version}"
+}
+
+#returns for example '7.4'
+function php_service_version () {
+    echo "$(ls /lib/systemd/system | grep php | grep fpm | sort | tail -1 | grep -o -P '(?<=php).*(?=-fpm)')"
```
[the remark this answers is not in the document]
liaralabs: Whatever is the most resilient. I imagine that the presence of the systemd services is actually a much better indicator of what's been installed by the system than the presence of the `/etc/php` folder and its contents as things can go wrong and that folder can disappear (something similar happened to Ubuntu recently but it was not exactly an easy fix)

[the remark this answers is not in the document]
flying-sausages: Let's stick with that then :)

PR #286, opened 2020-04-24
rm & apt functions, whiptail adjustments

flying-sausages:
In testing i noticed the source for libtorrent would fail to download and the previous version that was in place got compiled instead, so now that is being removed before.

Another function retrieves the repo version and displays it in the whiptails for deluge and rtorrent.

### DIFF
--- a/.github/auto-comment.yml
+++ b/.github/auto-comment.yml
@@ -1,0 +1,9 @@
+issueOpened: >
+  Thank you for raising a issue. We will try and get back to you as soon as possible. Please remember that this repository is maintained on volunteer time.
+
+  Please make sure you have given us as much context as possible, otherwise, we may not be able to help you.
+
+pullRequestOpened: >
+  Thank your for raising your pull request.
+
+  Please make sure you have followed our contributing guidelines. We will review it as soon as possible


sources/functions/short
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+#please make sure to give this full paths
+function rm_if_exists () {
+    path="$1"
+    if [[ -e $path ]]; then
+      rm -rf $path
```
userdocs, line 7: I think it might be sensible to quote the `$path` variable here in line 6 and 7, especially with a recursive `rm` in line 7.

The expansion of any path with spaces would fail. The use may ever present the problem but i think it would be good practice to quote a path expansion.

flying-sausages: Nice catch, that was my intention but it slipped me. Thanks!

sources/functions/short
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+#please make sure to give this full paths
+function rm_if_exists () {
+    path="$1"
+    if [[ -e "$path" ]]; then
+      rm -rf "$path"
+    fi
+}
+
```
flying-sausages, line 10: I wonder whether the filename for this is appropriate. I chose `short` as I intended this to store other small shortcuts.

liaralabs:
This will need to be rebased after the 20.04 merge. A function for rm_if_exists would be welcome -- you can toss it in the new utils function :smile:

PR #287, opened 2020-04-24
Allow to override default torrent client config values for directories

flying-sausages:
I am personally not a big fan of the enforced folder structure introduced by the scripts, and so I'd like to open possibilities for admins who create their users to override these without having to go and change something every time.

This will also make the configs' paths and directory creation easier to maintain as the default directory names are in one place, and so don't need to be cross-referenced and find&replaced.

If a user wants to change only one variable, they don't need to declare the others. If the file doesn't exist, no one notices and the result is the same.

I've gone ahead and mentioned this is unsupported in the code, but I'm more than happy to append the docs with this and check the issues for this ;)

If there's anything else we'd want to expose to the user, as a "`skel`-able" variable, this would be an optional path to pursue further.

### DIFF
--- a/scripts/install/deluge.sh
+++ b/scripts/install/deluge.sh
@@ -24,7 +24,6 @@ function _set_deluge_vars () {
 }
 
 function _dconf {
-  _set_deluge_vars
   for u in "${users[@]}"; do
     if [[ ${u} == ${master} ]]; then
       pass=$(cut -d: -f2 < /root/.master.info)
@@ -196,7 +195,6 @@ cat > /home/${u}/.config/deluge/hostlist.conf${SUFFIX} <<DHL
   ]
 }
 DHL
-  _set_deluge_vars
   echo "${u}:${pass}:10" > /home/${u}/.config/deluge/auth
   echo "localclient:${localpass}:10" >> /home/${u}/.config/deluge/auth
   chmod 600 /home/${u}/.config/deluge/auth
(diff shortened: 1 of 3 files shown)


flying-sausages:
Need to test the deluge code on a non-20.04 VM first, as the python scripts won't work there

flying-sausages:
deluge runs fine on 18.04, isn't supported on 20.04 yet.
rtorrent script runs fine on both 18.04 and 20.04.

scripts/install/deluge.sh
```diff
@@ -185,17 +196,17 @@ cat > /home/${u}/.config/deluge/hostlist.conf${SUFFIX} <<DHL
...
   chmod 600 /home/${u}/.config/deluge/auth
   chown -R ${u}.${u} /home/${u}/.config/
-  mkdir /home/${u}/dwatch
-  chown ${u}: /home/${u}/dwatch
-  mkdir -p /home/${u}/torrents/deluge
-  chown ${u}: /home/${u}/torrents/deluge
+  mkdir -p "/home/${u}/${watch_dir}"
+  chown ${u}: "/home/${u}/${watch_dir}"
+  mkdir -p "/home/${u}/$download_dir"
+  chown ${u}: "/home/${u}/${download_dir}"
```
flying-sausages, line 206: Should this also make and set owner for the other two directories when they're specified? will that happen automatically when the user enables them? If not, what about the watch directory, which isn't enabled by default?

flying-sausages:
@userdocs suggested an alternative solution to the problem through just allowing entire alternative configs. I'll be making that asap instead.

PR #293, opened 2020-04-25
Allow custom configs for rtorrent and deluge

flying-sausages:
This PR introduces a way to change the default config used for rtorrent and deluge.

It also achieves removing the actual client config out of the install scripts.

Additionally, an optional mechanism to create directories within userdirs is added so that this too can be done automated. I wanted to parse the configs but I thought that's a little over the top, and a user could also just make the dirs themselves if they so prefer. If there's a better way to do this please let me know.


flying-sausages:
- [ ] Fix deluge web failing

flying-sausages:
THe deluge web issues are possibly fixed when 28ddf371e24994b1933f6a2f5519a89045590ab2 is pushed, after restarting nginx and panel modules this seems to work fine.

flying-sausages:
This has been tested with rtorrent as well and works great.
I've used the function to make user directories to create them and all went smooth.

liaralabs:
The only issue I see right now is that the `.gitignore` isn't going to work quite as intended when using `box update` due to

```
function _update() {
  echo -e "Updating swizzin local repository"
  cd /etc/swizzin
  git checkout master
  git fetch origin master
  git reset --hard origin/master ## This line

  for f in /etc/swizzin/scripts/update/*; do 
    bash "$f"; 
  done
}
```

A hard reset will reset the *entire* repo to its clone state regardless of gitignores. It's quite annoying... Perhaps the hard reset can be removed; however the hard reset basically ensures that the git update works every time, removing the possibility of user error when a user updates scripts they shouldn't be touching (as this will break git pulls).

I worked around this in the upgrade script for the panel, but it's not exactly pretty

flying-sausages:
I think we should have a more general discussion on how we want non-whiptail customisability to work going further. This is one example.

The way Transmisssion was done would be another example. That relies on having env variables set before running the script. TO improve that, maybe we can add sourcing an .env file with a preset location like `.config/swzzin.env` to the box command where these variables can be stored for re-use? I'm not sure which is prettier, easier to maintain, and more useful... Some feedback/vision on this would be appreciated.

flying-sausages:
For some other time

PR #294, opened 2020-04-26
Transmission

flying-sausages:
This PR adds Transmission in a true multi-user fashion!

There is no nginx config just yet, however the RPC sockets are-a-workin!


flying-sausages:
I tried to work on the nginx proxy and could not get it working, so I revert to the last working version. Someone with a lot of nginx skills should take a look at it, let me know when you do so I can guide you through

scripts/install/transmission.sh
```diff
@@ -0,0 +1,189 @@
...
+
+    if [[ $incomplete_dir_enabled = "true" ]]; then 
+        mkdir -p /home/${user}/${incomplete_dir}
+        chown ${user}:${user} -R /home/${user}/${incomplete_dir}
+    fi
+}
+
+_mkconf_transmission () {
+    _setenv_transmission 
+cat > /home/${user}/.config/transmission-daemon/settings.json <<EOF
```
flying-sausages, line 66: Please check if any settings need to be changed here still

sources/functions/swizzin
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+function _get_user_password () {
+    user_to_get="$1"
+    password=''
+
+    ##TODO fix when there is no addition users other than .master
```
flying-sausages, line 7: need to still fix this

PR #295, opened 2020-04-26
Organizr

flying-sausages:
This Installs organizr as an app.

I based this off 20.04 as it has the nice PHP scripts.


flying-sausages:
- [ ] See if the first install can be bootstrapped

scripts/install/organizr.sh
```diff
@@ -0,0 +1,29 @@
...
+chown -R www-data:www-data /srv/organizr_db  
+
+##TODO make sure this is done right
+echo "Installing python requirements"
+pip install requests pprint  >> $log 2>&1
+echo "Bootstrapping Database"
+python /etc/swizzin/scripts/organizr.setup.py  >> $log 2>&1
+
+
+touch /install/.organizr.lock
```
flying-sausages, line 29: - [ ] Get this working

Help with this would be highly appreciated as I am not very good with python

flying-sausages: Python removed

flying-sausages:
Managed to succesfully curl against the form, however for some reason it is taking a good while for the user to be let through to the dashboard. Will try fix but as far as this PR goes I'm happy to merge this into a non-master branch for testing.

flying-sausages:
Issue here https://github.com/causefx/Organizr/issues/1370

causefx:
Nice write up.  As for the lag time, this makes zero sense to me as why it's happening.  All the api call does is make the default DB with the associated user inside it.  It then creates the config.php file as well.  The DB file is super small along with the new php file.  I will check the bash script out and will test locally to see what happens.  Also, so to clarify, after 10 mins or so, the user can login normally and the user will be presented with the settings page?

flying-sausages:
Hey CauseFX! It straight up works after approx 10 minutes...
TO add to this, I'm running this on a multipass VM with 1 core and 1GB ram. I don't see any process hanging from PHP or anything when this is waiting, so I really don't know where this is coming from.

causefx:
when you get a chance, can you run script and then check /api/config/config.php to see if file is created right after script run.  You can also check the if DB is created as well...

flying-sausages:
Just ran this again right now.
The file is created with the following content right after the script is done
```php
<?php
return array(
        'configVersion' => '2.0.408',
        'dbLocation' => '/srv/organizr_db/',
        'dbName' => 'db.db',
        'license' => 'personal',
        'organizrAPI' => 'P0XbyMNFdecIG7vP',
        'organizrHash' => 'FMDK5Lfsz6w77xJC',
        'registrationPassword' => '1gVaDHh1sSHfJkZ7'
);
```

The file exists too
```
ubuntu@moving-panther:/srv/organizr_db$ ls -lah
total 108K
drwxr-xr-x 2 www-data www-data 4.0K Apr 27 16:16 .
drwxr-xr-x 5 root     root     4.0K Apr 27 16:16 ..
-rw-r--r-- 1 www-data www-data  96K Apr 27 16:16 db.db
```

causefx:
navigate to organizr page but dont login - after that can you paste contents of config.php again?

On April 27, 2020 at 10:18 AM, flying-sausages <notifications@github.com> wrote:






Just ran this again right now.
The file is created with the following content right after the script is done

<?php
return array(
        'configVersion' => '2.0.408',
        'dbLocation' => '/srv/organizr_db/',
        'dbName' => 'db.db',
        'license' => 'personal',
        'organizrAPI' => 'P0XbyMNFdecIG7vP',
        'organizrHash' => 'FMDK5Lfsz6w77xJC',
        'registrationPassword' => '1gVaDHh1sSHfJkZ7'
);

The file exists too

ubuntu@moving-panther:/srv/organizr_db$ ls -lah
total 108K
drwxr-xr-x 2 www-data www-data 4.0K Apr 27 16:16 .
drwxr-xr-x 5 root     root     4.0K Apr 27 16:16 ..
-rw-r--r-- 1 www-data www-data  96K Apr 27 16:16 db.db

flying-sausages:
Before (ran script again)
```php
<?php
return array(
	'configVersion' => '2.0.408',
	'dbLocation' => '/srv/organizr_db/',
	'dbName' => 'db.db',
	'license' => 'personal',
	'organizrAPI' => '23weHpmx4EokYlWO',
	'organizrHash' => 'OLeaAvRODu9HPdLH',
	'registrationPassword' => 'V7b2mahwl7Czmyq5'
);
```
After
```php
<?php
return array(
	'configVersion' => '2.0.408',
	'dbLocation' => '/srv/organizr_db/',
	'dbName' => 'db.db',
	'license' => 'personal',
	'organizrAPI' => '23weHpmx4EokYlWO',
	'organizrHash' => 'OLeaAvRODu9HPdLH',
	'registrationPassword' => 'V7b2mahwl7Czmyq5',
	'uuid' => '6bdbaec0-2a5d-47a5-bb75-1aa6145d8cdf
);
```

flying-sausages:
After it starts working, the file does not seem to have any changes...
```php
<?php
return array(
	'configVersion' => '2.0.408',
	'dbLocation' => '/srv/organizr_db/',
	'dbName' => 'db.db',
	'license' => 'personal',
	'organizrAPI' => '23weHpmx4EokYlWO',
	'organizrHash' => 'OLeaAvRODu9HPdLH',
	'registrationPassword' => 'V7b2mahwl7Czmyq5',
	'uuid' => '6bdbaec0-2a5d-47a5-bb75-1aa6145d8cdf'
);'
```

causefx:
this boggles my mind...  I will have to test on my end...  But yeah, once those two files are created - there is nothing left for organizr to do.  It should allow for login.

flying-sausages:
Well, it _allows_ login, just doesn't let you further :P the cookies are set, correctly as far as I can see in the browser as well... When I log into the panel, I see that the session that was used to do the curl call is still registered there. Maybe something in how that is being handled?

flying-sausages:
I just tried to finish the setup manually and I am getting similar results with not being let through. So I suppose this is not in the curl but something else that's being done...

causefx:
can you take a gif or video file of the setup and once you try to login please?

On April 27, 2020 at 12:17 PM, flying-sausages <notifications@github.com> wrote:






I just tried to finish the setup manually and I am getting similar results with not being let through. So I suppose this is not in the curl but something else that's being done...

flying-sausages:
The roundup has been added to the issue on the organizr repo mentioned above. The user is currently asked to reload php-fpm manually.

causefx:
have you tried actually stopping the php service with the script and then starting with the script? instead of restarting?

flying-sausages:
@causefx I did, with a sleep of a second in there, and still nothing. Again, when it happens from the terminal directly and not from the script, it works fine...

We have a script for running php-cli to the fpm and ran this in the terminal 
`/usr/local/bin/swizzin/php-fpm-cli -r 'opcache_reset();` which fixed it, however it is yet again not effective when it happens from within the script.

PR #303, opened 2020-04-29
Webmin

flying-sausages:
Adds webmin to box!

Still need to see if I can disable the session auth and use basic http instead


flying-sausages:
The session being done through basic http auth can be done in the future, for now it is communicated to end user what account to use.

PR #320, opened 2020-05-01
Setdisk++

flying-sausages:
This introduces the following:
- passing arguments to setdisk to skip the `read` stuff if possible
- allowing use of `max` to give all space possible (to be used by server admins who don't want to see rutorrent bugs if no quota is set (Maybe that should be fixed too) )
- Add `setdisk` to the adduser procedure

### DIFF
--- a/scripts/setdisk
+++ b/scripts/setdisk
@@ -1,54 +1,81 @@
 #!/bin/bash
-#
-# [QuickBox Quota Set Script]
+# Swizzin Quota Set Script
+# Modified by flying_sausages from Quickbox files
 #
 # GitHub:   https://github.com/QuickBox/quickbox_packages
-# Author:   QuickBox.io
-# URL:      https://plaza.quickbox.io
-#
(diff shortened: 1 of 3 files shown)


scripts/setdisk
```diff
@@ -1,54 +1,84 @@
...
+fi
+case "$SIZE" in
+  max)
+    fstabid=$(grep quota /etc/fstab | awk '{printf $1}')
+    if [[ $fstabid =~ "UUID" ]]; then
+      uuid=$(echo "$fstabid" | cut -d= -f 2)
+      disk=$(blkid -U "$uuid")
+    elif [[ $fstabid =~ "LABEL" ]]; then
+      label=$(echo "$fstabid" | cut -d= -f 2)
+      disk=$(blkid -L "$label")
```
flying-sausages, line 44: Let me know if there could be any other types of values that should be handled

flying-sausages: blkid --help doesn't seem to suggest so

PR #324, opened 2020-05-12
 Remove default swizzin IRC references, improve user management

flying-sausages:


flying-sausages:
This also includes user management improvements through box as they were done without the CLI options thelounge exposes. Right now password is being passed via STDIN which is a bit hacky and so I opened this FR on thelounge https://github.com/thelounge/thelounge/issues/3913

PR #330, opened 2020-05-16
Improve nextcloud installer and implement user support

flying-sausages:
I tried it in my VM and it didn't work so i escaped a bunch of things and now it works. I also didn't like how nonverbose it was and though it died so i added some echos. Also no need for mysql root user password if you are root user so out with that. Oh and I added all the `occ` actions to adduser, deluser, chpass etc so that nextcloud users are managed by box too.


flying-sausages:
- [ ] Setup database with utf8mb4
- [ ] Fix the indicies that are missing and stuff

flying-sausages:
I can't figure out how to satisfy the requirements for the `utf8mb4` from nextcloud, but I suppose it's fine if that's what it is. These are the messages I still see after the installer runs:

image

flying-sausages:
This needs to be tested on Debian for sure.
We need to decide if we want emoji support or not as I'm happy without it; after all it's been like this forever and we never heard anything.
Do we want to silence the nginx config and opcache stuff too?

flying-sausages:
Very ready

flying-sausages:
Revisiting in #497 due to broken tree and old fork

PR #341, opened 2020-05-27
Mylar

Public920:
Mylar is an automatic Comic Book downloader based on Headphones
https://github.com/mylar3/mylar3


scripts/install/mylar.sh
```diff
@@ -0,0 +1,72 @@
...
+    curl -O https://www.python.org/ftp/python/3.8.1/Python-3.8.1.tar.xz
+    tar -xf Python-3.8.1.tar.xz
+    cd Python-3.8.1
+    make
+    make install
+
+    python3_getpip
+fi
+
+python3_venv ${user} mylar
```
flying-sausages, line 37: Please take a look at how some of the other apps are being installed using the already defined and shared python venving functions. If there is something here that is re-usable, please make a function for it as well.

scripts/install/mylar.sh
```diff
@@ -0,0 +1,72 @@
...
+Wants=network.target network-online.target
+After=network.target network-online.target
+
+[Service]
+Type=forking
+User=${user}
+Group=${user}
+ExecStart=/opt/.venv/mylar/bin/python3 /opt/mylar/Mylar.py -d --pidfile /run/${user}/mylar.pid --datadir /opt/mylar --nolaunch --config /opt/mylar/config.ini --port 8090
+PIDFile=/run/${user}/mylar.pid
+
```
flying-sausages, line 57: No restart on failure? Maybe it's a valid choice, I'm just curious.

scripts/install/mylar.sh
```diff
@@ -0,0 +1,72 @@
...
+
+[Install]
+WantedBy=multi-user.target
+MYLRSD
+
+systemctl enable --now mylar >>$log 2>&1
+sleep 10
```
flying-sausages, line 64: What is this required for?

scripts/nginx/mylar.sh
```diff
@@ -0,0 +1,430 @@
...
+    cat > /etc/nginx/apps/mylar.conf <<MYNG
+location /mylar {
+    include /etc/nginx/snippets/proxy.conf;
+    proxy_pass http://127.0.0.1:8090/mylar;
+    auth_basic "What's the password?";
+    auth_basic_user_file /etc/htpasswd.d/htpasswd.${user};
+}
+MYNG
+fi
+cat > /opt/mylar/config.ini <<MYCONF
```
flying-sausages, line 22: Seeing as this is in the nginx folder, this will only get triggered if:
User has nginx installed before the app
User installs nginx after the app has been installed.

THerefore I'm not sure this is the place to cat out a config that's specifically for `mylar` and has not much to do with the nginx. If you don't have nginx, you won't get this config file as it is right now.

scripts/nginx/mylar.sh
```diff
@@ -0,0 +1,430 @@
...
+transmission_directory = None
+
+[Deluge]
+deluge_host = None
+deluge_username = None
+deluge_password = None
+deluge_label = None
+deluge_pause = False
+deluge_download_directory = ""
+deluge_done_directory = ""
```
flying-sausages, line 409: Would be sweet if these could be pre-configured, just a thought :)

scripts/install/mylar.sh
```diff
@@ -0,0 +1,72 @@
...
+else
+    LIST='git build-essential zlib1g-dev libncurses5-dev libgdbm-dev libnss3-dev libssl-dev libreadline-dev libffi-dev curl libbz2-dev'
+fi
+
+apt-get -y update >>"$log" 2>&1
+for depend in $LIST; do
+    apt-get -qq -y install $depend >>"$log" 2>&1 || { echo "ERROR: APT-GET could not install a required package: ${depend}. That's probably not good..."; }
+done
+
+if [[ ! $codename =~ ("buster"|"focal") ]]; then
```
flying-sausages, line 26: We'd like to avoid specifying lists that will only have to keep growing and be maintained, and prefer to invert them so that we can tick off the OSs that we'll be dropping eventually.
`$codename =~ ("buster"|"focal")` should be `! $codename =~ (xenial|jessie|bionic)` or the appropriate form of this idea.

scripts/nginx/mylar.sh
```diff
@@ -0,0 +1,430 @@
...
+fi
+
+if [[ ! -f /etc/nginx/apps/mylar.conf ]]; then
+    cat > /etc/nginx/apps/mylar.conf <<MYNG
+location /mylar {
+    include /etc/nginx/snippets/proxy.conf;
+    proxy_pass http://127.0.0.1:8090/mylar;
+    auth_basic "What's the password?";
+    auth_basic_user_file /etc/htpasswd.d/htpasswd.${user};
+}
```
flying-sausages, line 19: you should be also reloading nginx by triggering `systemctl reload nginx` at some point so that the config is applied.

scripts/install/mylar.sh
flying-sausages: i.e. From my limited knowledge of how python apps are being handled right now, we only venv things when python2 is needed and we let the python3 apps be handled normally unless something specific is needed.

scripts/install/mylar.sh
```diff
@@ -0,0 +1,72 @@
...
+    cd /tmp
+    curl -O https://www.python.org/ftp/python/3.8.1/Python-3.8.1.tar.xz
```
flying-sausages, line 28: Not sure it's a good idea to hard-code the version of python for this. Is there a specific requirement for 3.8.1?

scripts/install/mylar.sh
```diff
@@ -0,0 +1,72 @@
...
+
+if [[ $codename =~ ("buster"|"focal") ]]; then
+    LIST='git python3-dev virtualenv python3-virtualenv python3-pip'
+else
+    LIST='git build-essential zlib1g-dev libncurses5-dev libgdbm-dev libnss3-dev libssl-dev libreadline-dev libffi-dev curl libbz2-dev'
+fi
+
+apt-get -y update >>"$log" 2>&1
+for depend in $LIST; do
+    apt-get -qq -y install $depend >>"$log" 2>&1 || { echo "ERROR: APT-GET could not install a required package: ${depend}. That's probably not good..."; }
```
flying-sausages, line 23: please make sure to add `export DEBIANFRONTEND=noninteractive` or the right form of that command, and also the DPKG options that will use old configs. A nice example is in the quota installer that I can think of off the top of my head.

scripts/nginx/mylar.sh
```diff
@@ -0,0 +1,430 @@
...
+https_chain = None
+https_force_on = False
+host_return = None
+authentication = 0
+login_timeout = 43800
+alphaindex = True
+
+[API]
+api_enabled = False
+api_key = None
```
flying-sausages, line 125: Might want to pre-fill this with some generated value so that we can then plug this into whatever else will support it later

scripts/nginx/mylar.sh
```diff
@@ -0,0 +1,430 @@
...
+weekfolder = False
+weekfolder_loc = None
+weekfolder_format = 0
+indie_pub = 75
+biggie_pub = 55
+pack_0day_watchlist_only = True
+reset_pullist_pagination = True
+
+[Interface]
+http_port = 8090
```
flying-sausages, line 108: I would double-check no other swizzin app has the same port configured

scripts/nginx/mylar.sh
flying-sausages, line 22: Another remark, I think it would be nice to change these settings to whatever the most obvious choices for swizzin users would be. Feel free to make these deviate from the standard config, and if possible leave a comment above/around the modifications so we know where to look in case someone comes for some support questions.

scripts/nginx/mylar.sh
```diff
@@ -0,0 +1,430 @@
...
+    proxy_pass http://127.0.0.1:8090/mylar;
+    auth_basic "What's the password?";
+    auth_basic_user_file /etc/htpasswd.d/htpasswd.${user};
+}
+MYNG
+fi
+cat > /opt/mylar/config.ini <<MYCONF
+[General]
+oldconfig_version = None
+auto_update = False
```
flying-sausages, line 25: porque no los auto-updatos? :stuck_out_tongue:

scripts/nginx/mylar.sh
```diff
@@ -0,0 +1,430 @@
...
+http_username = None
+http_password = None
+http_root = /mylar
+enable_https = False
+https_cert = /opt/mylar/server.crt
+https_key = /opt/mylar/server.key
+https_chain = None
+https_force_on = False
+host_return = None
+authentication = 0
```
flying-sausages, line 119: If there's a chance to use the basic auth for this, might be super useful

scripts/nginx/mylar.sh
```diff
@@ -0,0 +1,430 @@
...
+
+[Logs]
+log_dir = /opt/mylar/logs
+max_logsize = 10000000
+max_logfiles = 5
+log_level = 1
+
+[Git]
+git_path = None
+git_user = mylar3
```
flying-sausages, line 144: What is this setting for? has this functionality been tested? we don't have a myla3 user being created on the system but I'm not sure that's what it is.

scripts/nginx/mylar.sh
```diff
@@ -0,0 +1,430 @@
...
+search_storyarcs = False
+
+[Metatagging]
+enable_meta = False
+cmtagger_path = None
+cbr2cbz_only = False
+ct_tag_cr = True
+ct_tag_cbl = True
+ct_cbz_overwrite = False
+unrar_cmd = None
```
flying-sausages, line 324: might as well configure these, unrar should be installed already iirc

scripts/install/mylar.sh
```diff
@@ -0,0 +1,72 @@
...
+    cd Python-3.8.1
+    make
+    make install
+
+    python3_getpip
+fi
+
+python3_venv ${user} mylar
+
+git clone https://github.com/mylar3/mylar3.git /opt/mylar >>"$log" 2>&1
```
flying-sausages, line 39: I would consider pointing this to the releases instead as well. It would be a lot easier to support this app if we can stick to this system, also it will probably mean the updating won't be "broken" after an install. I'm doing something similar in the mango PR.

scripts/install/mylar.sh
```diff
@@ -0,0 +1,472 @@
...
+if [[ -f /tmp/.install.lock ]]; then
+    log="/root/logs/install.log"
+else
+    log="/root/logs/swizzin.log"
+fi
+
+user=$(cut -d: -f1 < /root/.master.info)
+password=$(cut -d: -f2 < /root/.master.info)
+codename=$(lsb_release -cs)
+
```
flying-sausages, line 14: `echo "Setting up python environment"`

scripts/install/mylar.sh
```diff
@@ -0,0 +1,472 @@
...
+    pyenv_install_version 3.8.1
+    pyenv_create_venv 3.8.1 /opt/.venv/mylar
+else
+    export DEBIAN_FRONTEND=noninteractive
+    apt-get -y update >> $log 2>&1
+    apt-get -q -y install python3-dev python3-pip python3-venv >> $log 2>&1
+    mkdir -p /opt/.venv/mylar
+    python3 -m venv /opt/.venv/mylar
+fi
+
```
flying-sausages, line 27: `echo "Downloading Mylar source"`

scripts/install/mylar.sh
```diff
@@ -0,0 +1,472 @@
...
+mkdir /opt/mylar
+curl -s https://api.github.com/repos/mylar3/mylar3/releases/latest \
+    | grep tarball_url \
+    | cut -d '"' -f 4 \
+    | tr -d \" \
+    | xargs -n 1 curl -sSL \
+    | tar -xz -C /opt/mylar --strip-components=1  >> $log 2>&1
+/opt/.venv/mylar/bin/pip3 install -r /opt/mylar/requirements.txt >> $log 2>&1
```
flying-sausages, line 35: `echo "Satisfying mylar dependencies"`

scripts/install/mylar.sh
```diff
@@ -0,0 +1,472 @@
...
+
+chown -R ${user}: /opt/mylar
+chown -R ${user}: /opt/.venv/mylar
+
+cat > /opt/mylar/config.ini <<MYCONF
```
flying-sausages, line 40: `echo "Generating Mylar configuration and service files"`

scripts/install/mylar.sh
```diff
@@ -0,0 +1,472 @@
...
+WantedBy=multi-user.target
+MYLRSD
+
+systemctl daemon-reload >> $log 2>&1
+systemctl enable --now mylar >> $log 2>&1
+
+if [[ -f /install/.nginx.lock ]]; then
+    bash /usr/local/bin/swizzin/nginx/mylar.sh
+    systemctl reload nginx
+    echo "Install complete! Please note Mylar access url is: https://$(ip route get 1 | sed -n 's/^.*src \([0-9.]*\) .*$/\1/p')/mylar/home"
```
flying-sausages, line 469: Might want to also figure out how this would work for people with domains

scripts/nginx/mylar.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# Nginx Configuration for Mylar
+# Author: Public920
+
+user=$(cut -d: -f1 < /root/.master.info)
+
+isactive=$(systemctl is-active mylar)
+if [[ $isactive == "active" ]]; then
+    systemctl stop mylar
```
flying-sausages, line 9: Is this necessary if there is no configuration being made to mylar itself? Might be a remnant of when the config was still here.

flying-sausages:
@Public920 is this ready? I can't remember what the state of this was or whether there were some changes we talked about that still could have had some improvement

Public920:
I'll take a look at my local to check if there's any changes I've made that can be pushed, but from memory the core of it all works fine. The only things left were some quality of life changes to do with presetting config and info output.

flying-sausages:
> I'll take a look at my local to check if there's any changes I've made that can be pushed, but from memory the core of it all works fine. The only things left were some quality of life changes to do with presetting config and info output.

Any updates?

flying-sausages:
@Public920 Let me know if you need some help making this align with the new changes in the repo. Otherwise I will close this in a bit for inactivity.

flying-sausages:
Please re-open whenever you'd like to revisit this. A lot has happened since, so a lot of new feedback would be necessary for sure ;)

wishx:
Note that feature development has stopped as we have moved to Mylar3.
EOL for this project is the end of 2020 and will no longer be supported. This repo will be archived as of Jan-1-2021.
https://github.com/mylar3/mylar3

PR #347, opened 2020-06-01
Fix letsencrypt failure if Cloudflare domain is on allowed TLD but has substring of a prohibited TLD

epicaric:
The current regex used to prohibit TLDs that Cloudflare doesn't support is wrong, it should look for Anything.ProhibitedTLD but instead it looks for AnythingProhibitedTLDAnything. 

For example, the domain "the-abbreviation-for-the-state-of-georgia-is-ga.com" will be rejected because it contains some characters, followed by "ga" (a prohibited TLD), followed by some characters.

This change ensures that the regex only matches against a literal period, then a prohibited TLD, then the end of the hostname.

### DIFF
--- a/scripts/install/letsencrypt.sh
+++ b/scripts/install/letsencrypt.sh
@@ -49,7 +49,7 @@ esac
 
 if [[ ${cf} == yes ]]; then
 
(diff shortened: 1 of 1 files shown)


flying-sausages:
Oh yeah absolutely, that was something I introduced, my bad.

PR #354, opened 2020-06-29
Sonarr v3, [y/n] function

flying-sausages:
This PR introduces installation for v3, as well as migration steps in case `sonarr` is installed.

A function to simplify and unify [y/n] prompts has been introduced

~~Also, a method for extra dependency installation is introduced.~~


flying-sausages:
Documentation here https://github.com/liaralabs/docs.swizzin.ltd/pull/21

flying-sausages:
blocked by #361

flying-sausages:
This requires jq to be available in the system. A second PR should be made to convert the rest of swizzin to use `ask` and then implement coloured echos into it too.

scripts/install/sonarrv3.sh
```diff
@@ -0,0 +1,139 @@
...
+        #shellcheck source=sources/functions/ask
+        . /etc/swizzin/sources/functions/ask
+
+        if ! ask "Do you want to continue?" N; then
+            exit 0
+        fi
+
+        # TODO make backup
+        # TODO 
+        echo "Backing up Sonarr v2" | tee -a $log
```
flying-sausages, line 35: Remove TODOs as it's done

scripts/install/sonarrv3.sh
```diff
@@ -0,0 +1,139 @@
...
+        echo "Sonarr was not found from apt.sonarr.tv repository. Please inspect the logs and try again later."
+        exit 1
+    fi
+
+}
+
+_install_sonarrv3 () {
+    echo "Installing Sonarr v3 from apt" | tee -a $log
+    # settings relevant from https://github.com/Sonarr/Sonarr/blob/phantom-develop/distribution/debian/config
+    master=$(cut -d: -f1 < /root/.master.info)
```
flying-sausages, line 105: Maybe environment variable for non-master user?

scripts/install/sonarrv3.sh
```diff
@@ -0,0 +1,139 @@
...
+            echo "We cannot trigger Sonarr to dump a current backup, but the current files and previous weekly backups can still be copied" | tee -a $log
+            if ! ask "Continue without triggering internal Sonarr backup?" N; then 
+                exit 1
+            fi
+        else
+            echo "Sonarr backup Job ID = $id, waiting to finish" >> $log
+
+            status=""
+            while [[ $status != "\"completed\"" ]]; do 
+                status=$(curl -s "${address}/command/$id?apikey=${apikey}" | jq '.status')
```
flying-sausages, line 55: Could fix the `"\"` business with the `jq` `raw` flag

flying-sausages:
Should I also implement a query which will ask a user which usergroups should have sonarr added? i.e. `usermod -a -G $user sonarr`, together with making Sonarr run as its own user instead?

This would seem like a cleaner solution in the long run, would cause less confusion with permissions and should be easier to troubleshoot in the future.

flying-sausages:
Sonarrv3 will now run as a system user as opposed to the master user. Master user will have the sonarr user added to its usergroup, and other users' directories can be added to this as part of the installation process

flying-sausages:
Tested and working as expected
- Install on focal (with & without nginx)
- removal on focal
- attempting to install v2 over v3 and getting blocked
- upgrade from nzbdrone to sonarr (with & without nginx) on focal and migration
- FS permissions allowing sonarr to function

To test:
- Debian install? Not sure what would be out of line...

To implement:
- Nothing

scripts/install/sonarrv3.sh
flying-sausages: whatever

flying-sausages:
- [ ] change `localhost` to `127.0.0.1`

flying-sausages:
- [ ] Add `ask` to make backup

scripts/install/sonarr.sh
```diff
@@ -19,7 +19,39 @@
...
+  if [[ $v3present == "true" ]]; then
+    echo "Sonarr v3 detected. If you want to proceed installing Sonarr v2, please remove Sonarr v3 first."
+    exit 1
+  fi
+
+  if [[ -d /root/sonarrv2.bak ]]; then 
+    echo
+    echo "WARNING: Found backups before Sonarr v3 installation"
+    echo "Follow these steps post-install https://github.com/Sonarr/Sonarr/wiki/Backup-and-Restore"
+    echo "The backup is stored in /root/sonarrv2.bak/"
```
liaralabs, line 40: Can we put the backup somewhere other than just dumping it in the root dir? I find this folder can easily get cluttered.

I know it's not the best, but currently the deluge backup is stored right next to the deluge directory in .config. Otherwise, a standardized backup location?

flying-sausages: Sure thing, where should it go?

scripts/install/sonarr.sh
```diff
@@ -19,7 +19,39 @@
...
+    #TODO implement restore procedure if user wants that to happen?
+    #TODO check if unit is still masked. Unmaks it if that's the case.
+  fi
+
+}
+
 function _installSonarrintro() {
+  echo "Sonarr v3 is available as a separate application (sonarrv3)"
+  . /etc/swizzin/sources/functions/ask
+  if ! ask "Would you like to continue installing Sonarr v2?"; then
```
liaralabs, line 51: Does this prompt a question during setup? Ideally this should be avoided because we have to assume the user specifically chose v2 because v3 would be listed right beneath it.

scripts/install/sonarrv3.sh
```diff
@@ -0,0 +1,179 @@
...
+
+        if ask "Would you like to trigger a Sonarr-side backup?" Y; then
+            echo "Backing up Sonarr v2" | tee -a $log
+            if [[ -f /install/.nginx ]]; then
+                address="http://127.0.0.1:8989/sonarr/api"
+            else
+                address="http://127.0.0.1:8989/api"
+            fi
+
+            [[ -z $sonarrv2owner ]] && sonarrv2owner=$(cut -d: -f1 < /root/.master.info)
```
liaralabs, line 41: Was this written before `_get_master_username`?

scripts/install/sonarr.sh
flying-sausages: I'm assuming users won't know `sonarrv3` is even available. I never check the list, and we have no way to really inform people that new packages are rolling out. It's probablt not gonna land on the swizzin.tld main site anytime soon either, which I assume gets a lot more traffic than the docs or the github commits.

scripts/install/sonarrv3.sh
```diff
@@ -0,0 +1,179 @@
...
+            if [[ ! -d /home/"${sonarrv2owner}"/.config/NzbDrone ]]; then
+                echo "No Sonarr config folder found for $sonarrv2owner. Exiting" | tee -a $log
+                exit 1
```
liaralabs, line 44: Is exiting the right thing to do here?

If sonarr is installed but the config cannot be found, should we simply just instruct users to uninstall v2 and exit or just ask them if we want to continue anyway? Otherwise we could give them an actionable item such as:

"No Sonarr config folder found for /home/"${sonarrv2owner}"/.config/NzbDrone. Does it exist?"

Perhaps a `read -p` could be used for the user to confirm that continuing may not have the results expected or that they should quit the script themselves now.

scripts/install/sonarrv3.sh
flying-sausages: Good catch! I think it was, I'll switch it out tomorrow

scripts/install/sonarrv3.sh
flying-sausages: They could just re-start the installation without the migration too, and we do tell them about the docs at the beginning so I suppose that's the first thing they'd check. I'd rather do that than give them a way to do a botched migration.

scripts/install/sonarrv3.sh
```diff
@@ -0,0 +1,179 @@
...
+                echo "Failure triggering backup (see logs)." | tee -a $log
+                echo "We cannot trigger Sonarr to dump a current backup, but the current files and previous weekly backups can still be copied" | tee -a $log
+                if ! ask "Continue without triggering internal Sonarr backup?" N; then
+                    exit 1
+                fi
+            else
+                echo "Sonarr backup Job ID = $id, waiting to finish" >> $log
+
+                status=""
+                while [[ $status != "\"completed\"" ]]; do
```
liaralabs, line 66: If the backup fails this will create an infinite loop.

Does sonarr assign a pending status for tasks in progress?

flying-sausages: It sets it as "running" or something while it is not finished. When it is, the API will return that. If it creates an infinite loop, the user will know that it's just the backup that failed and it won't make a huge issue with the installation. Could add some timeout or something if you'd prefer. How about 20 seconds?

scripts/install/sonarrv3.sh
```diff
@@ -0,0 +1,179 @@
...
+                usermod -a -G "$u" "$sonarrv3owner" | tee -a $log
+                chmod g+rwx /home/"$u"
+            done
+        fi
+}
+
+_nginx_sonarr () {
+    if [[ -f /install/.nginx.lock ]]; then
+        #TODO what is this sleep here for? See if this can be fixed by doing a check for whatever it needs to
+        sleep 20
```
liaralabs, line 155: This was to make sure the config was properly populated before we attempt to manipulate it for nginx.

Sonarr can take a bit to start sometimes in my experience which is why I wanted to make sure it had enough time. If the config can just be `cat`'d in this can probably be ignored.

scripts/nginx/sonarrv3.sh
```diff
@@ -0,0 +1,44 @@
...
+}
+SONARR
+
+isactive=$(systemctl is-active sonarr)
+
+if [[ $isactive == "active" ]]; then
+  systemctl stop sonarr
+fi
+
+cat > /var/lib/sonarr/config.xml <<SONN
```
liaralabs, line 24: I do not think we should be using `/var/lib/sonarr` for sonarr storage. I know this is default but nzbdrone storage has always been held in .config. `~/.config` already contains the data of the majority of applications and I don't really want to move apps that are already there out of there.

scripts/install/sonarrv3.sh
flying-sausages: `cat`'d in? Not sure I understand what that would achieve. Do you mean grepping for some content in a while loop?

scripts/remove/sonarr.sh
```diff
@@ -18,21 +18,24 @@
 #   under the GPL along with build & install instructions.
 #
 username=$(cut -d: -f1 < /root/.master.info)
-local_setup=/etc/QuickBox/setup/
```
liaralabs, line 21: :sweat_smile:

scripts/nginx/sonarrv3.sh
flying-sausages: This should do it, leaving this here for later https://github.com/Sonarr/Sonarr/blob/phantom-develop/distribution/debian/config#L11

But I'd say I'd rather leave this as unmodified as possible given it's in the package manager. It's all documented too. Let me know if you won't budge on this and I'll just change it though.

scripts/install/sonarrv3.sh
flying-sausages: Added 20s counter, should do it

scripts/install/sonarr.sh
liaralabs: > It's probablt not gonna land on the swizzin.tld main site

:facepalm: 

I just want to ensure minimal questions are asked during setup to make it as unattended as possible once it starts installing -- I have been doing my best to streamline and automate things where possible (see pyload) to remove user interaction for setting things up, maybe it would be possible to ask this question earlier?

Plex is a notable exception to this rule, but I generally don't install it on the first pass anymore

scripts/install/sonarr.sh
liaralabs: `~/.config` is already being used, otherwise, `/root/backups` or if we intend to further extend in the future `/root/swizzin/backups`

scripts/install/sonarrv3.sh
liaralabs: I don't know if 20 seconds would be enough for some super huge libraries out there. Isn't it tarring up zillions of tiny files?

Can we not check to see if the status is `running` and then once it's no longer running we catch the response and handle it?

scripts/install/sonarrv3.sh
flying-sausages: Oh I see what you mean. I'd rather let it populate itself as this might change over time and I don't want to maintain that, plus preserving an API key if any is gonna be a pain. We could instead use `xq` which is `jq` for xml though, and add that value in through that. then again, if sonarr does it, might as well wait

liaralabs: Like if we are injecting the file to be pre-filled rather than waiting for it to be initialized then making changes with `sed`

scripts/nginx/sonarrv3.sh
liaralabs: Perhaps we should open it up to community review as I do not think that changing an established config location is a great idea. People are already used to configs being stored here, I just don't see the reason to change this.

liaralabs (commented):
Didn't mean to make a full review, but here it is

scripts/install/sonarr.sh
flying-sausages: > maybe it would be possible to ask this question earlier?

It should be the very first thing that gets ran, it's just that the functions are ordered weird.

scripts/install/sonarrv3.sh
flying-sausages: Oh I see what you mean, got it. I'll test it out tomorrow. There could be some other status responses other than `completed` you are right.

scripts/nginx/sonarrv3.sh
flying-sausages: I would say it's not getting changed as it's a new package completely. We could maybe symlink it in there for convenience's sake..?

scripts/install/sonarrv3.sh
liaralabs: Yeah, sorry, that's what I was trying to convey. If the task returns a status other than completed, but it finished then we should handle that a bit more gracefully than purgatory :smile:

scripts/install/sonarr.sh
flying-sausages: I like `/root/swizzin/` and `backups` in there, I'll go with that.

scripts/install/sonarrv3.sh
flying-sausages: Tested and resolved

scripts/install/sonarrv3.sh
flying-sausages: I'd leave this in seeing as the config needs to be present to set up the db and all from what I hear, will look at this if someone has an issue with it down the road

scripts/nginx/sonarrv3.sh
flying-sausages: Changed to ~/.config

scripts/install/sonarr.sh
flying-sausages: I caved in and disabled this.

PR #361, opened 2020-07-11
Apt functions and global dependency updater

flying-sausages:
Functions to wrap `apt` functions like `update`, `upgrade` and `install` that will have standard behaviour and command line options to streamline the process across all application installations

+ An update script for dependencies will allow us to install global packages which aren't application bound (e.g. `jq`)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,9 @@
+{
+    "cSpell.words": [
+        "dpkg",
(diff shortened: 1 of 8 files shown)


flying-sausages:
This one is ready for review and a merge into a non-master branch. When we're happy with the way it works, a new PR should be made for the implementation in installer/remover scripts

PR #362, opened 2020-07-11
HOTFIX: Transmission watchdirs and function fixes

flying-sausages:
Blocked by #361 

Environment variables would not re-set when making multi-user installs, so RPC ports are unset between installs.
Watchdirs are now also created and enabled by default.


scripts/nginx/transmission.sh
```diff
@@ -32,8 +32,16 @@ for u in ${users[@]}; do
     if [[ $active == "active" ]]; then
         systemctl stop transmission@${u}
     fi
-    sed -i 's/"rpc-bind-address": "0.0.0.0",/"rpc-bind-address": "127.0.0.1",/g' /home/${u}/.config/transmission-daemon/settings.json
-    RPCPORT=$(grep rpc-port /home/${u}/.config/transmission-daemon/settings.json | grep -o '[[:digit:]]*')
+
+    while systemctl is-active transmission@"$u" > /dev/null ;do
+        echo "is active"
```
flying-sausages, line 37: this needs to go

PR #364, opened 2020-07-14
Add code to generate qr code pngs to display in panel

flying-sausages:
Liara pls add wg qrcodes to panel

### DIFF
--- a/scripts/install/panel.sh
+++ b/scripts/install/panel.sh
@@ -83,3 +83,14 @@ EOSUD
 systemctl enable --now panel > ${log} 2>&1
 
 touch /install/.panel.lock
+
+# Ensure permissions for wgqr.png are correct if the file exists
+if [[ -f /install/.wireguard.lock ]]; then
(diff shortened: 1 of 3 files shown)


flying-sausages:
- [ ] TODO: ACLs for swizzin to read the config

flying-sausages:
image

flying-sausages:
I'm fine with not doing this, the docs are clear enough now on how anyone can do this anyway. Maybe something for down the road

liaralabs:
To be clear, I'm not against this, but my concern is trying to balance some people's fear that the panel may have too many permissions and that the QR code contains private keys. On a multi-user server this amounts to the swizzin user having potential access to any of the wg tunnels.

We already set the ACLs so this is generally a larger question about how permissive we are okay with the swizzin user being.

flying-sausages:
Yeah no worries, I don't have the bandwidth to make suggestions on this either rn. I'll move it to feathub

PR #399, opened 2020-09-11
APT Update: merge master and implement in conflicts

flying-sausages:


flying-sausages:
There is an issue with installing deluge but I believe that is a regression on focal, as the installer is asking for packages not available in focal.

A hotfix for that should go into production asap

```
ubuntu@fusionvm:~$ sudo box install deluge
Installing deluge 
/etc/swizzin/sources/functions/libtorrent: line 5: pkg-config: command not found
Building libtorrent-rasterbar
E: Unable to locate package python-libtorrent
Error: The simulation check resulted in an error. Please consult the log. Aborting install
```

PR #401, opened 2020-09-14
feat: Calibre, Content server and Calibre-web

flying-sausages:
## Description

Introduces the calibre stack together with calibre-web

## Fixes issues: 
N/A

## Proposed Changes:
- Added following applications
   - `calibre`
   - `calibre-cs`
   - `calibre-web` 

## Change Categories

- New feature 

## Checklist

    - PR link: https://github.com/liaralabs/docs.swizzin.ltd/pull/27
    - PR link: https://github.com/liaralabs/swizzin_dashboard/pull/30

### Architectures

| Focal 		|		✅	|			|	✅		|		❎ 		|
| Bionic		|			|			|			|				|
| Buster		|	??		|			|			|				|
| Stretch		|			|			|			|				|
| Raspbian  	|	⚫️		|	??		|	⚫️		|	⚫️			|

### ✅❎ Passed
- `calibre`
   -  Fresh install (not MU or with a web-app)
   - Upgrader
- `calibre-cs`
   - Fresh install with one user
   - Fresh install with and without nginx  
### 🛠🛠 TODO
- `calibre`
   - build from source on arm due to packages being way too out of date
- `calibre-cs`
   - Decision on MU status, calibre-based auth, and user management
   - Implement User management when this releases https://github.com/kovidgoyal/calibre/commit/0727fb7793c6369ef1f650ffe9873d2da722636c
   - Add panel entries for calibre content server
- `calibre-web`
   -  Further test why sometimes library fails to set, and wait to see if this can be used https://github.com/janeczku/calibre-web/issues/1817
### ❌❌ Currently failing
- `calibre-web`
   - user management is essentially out of scope due to the tooling for the time being
   - Changing the admin password is randomly failing on some systems https://github.com/janeczku/calibre-web/issues/1824


scripts/install/calibre-web.sh
```diff
@@ -0,0 +1,109 @@
...
+        calibredbuser=$(cut -d: -f1 < /root/.master.info)
+    fi
+    usermod -a -G "${calibredbuser}" $calibrewebuser >> $log 2>&1
+
+    sudo chown -R $calibrewebuser:$calibrewebuser $calibrewebdir
+
+    echo "Installing python dependencies" | tee -a $log
+    runuser -l $calibrewebuser -c "python3 -m pip install --target vendor -r $calibrewebdir/requirements.txt --upgrade" >> $log 2>&1
+    #fuck ldap. all my homies hate ldap
+    sed '/ldap/Id' -i $calibrewebdir/optional-requirements.txt
```
chadberg, line 50: Yeah, for the best. You want LDAP? They you know enough to know what you are doing and can do it yourself. Leave that stank out of this script.

flying-sausages:
Panel entry: https://github.com/liaralabs/swizzin_dashboard/pull/18
Docs: https://github.com/liaralabs/docs.swizzin.ltd/pull/27

flying-sausages:
Ready

scripts/install/calibre-web.sh
```diff
@@ -0,0 +1,124 @@
...
+calibrewebuser="calibreweb" # or make this master user?
+codename=$(lsb_release -cs)
+
+function _install_dependencies_calibreweb() {
+	apt_install unzip imagemagick
+}
+
+function _install_calibreweb() {
+	echo_progress_start "Creating venv for calibre-web"
+	if [[ $codename =~ ("bionic"|"stretch"|"xenial") ]]; then
```
liaralabs, line 15: Can you just add a comment regarding the minimum python version required? At some point I'm going to update the logic of the python venvs to use the python version rather than relying on codename

scripts/install/calibre-web.sh
```diff
@@ -0,0 +1,124 @@
...
+WantedBy=multi-user.target
+EOF
+	systemctl daemon-reload -q 2>&1 | tee -a $log
+	systemctl enable -q --now calibre-web.service 2>&1 | tee -a $log
+	echo_progress_done
+}
+
+_postinstall_calibreweb() {
+	echo_warn "Please continue the configuration and installation of Calibre-web through your browser."
+	hostname="<domain/IP>" #figure this one out dynamically maybe?
```
liaralabs, line 106: We have the IP command:

```
ip=$(ip route get 1 | sed -n 's/^.*src \([0-9.]*\) .*$/\1/p')
```

Hostname is tough because I'd warrant a guess that no one is setting up their hosts correctly to use something like `hostname -f`

scripts/remove/calibre-web.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+if [[ -f /tmp/.install.lock ]]; then
+	log="/root/logs/install.log"
+else
+	log="/root/logs/swizzin.log"
+fi
```
liaralabs, line 7: Outdated :smile:

scripts/install/calibre-web.sh
flying-sausages: Just says "Install dependencies by running pip3 install --target vendor -r requirements.txt (python3.x) or pip install --target vendor -r requirements.txt (python2.7)."

flying-sausages:
```suggestion
    # Minimum version is Python 3.x
	if [[ $codename =~ ("bionic"|"stretch"|"xenial") ]]; then
```

scripts/install/calibre-web.sh
flying-sausages: I'm gonna just do away with it completely and change the wording

scripts/remove/calibre-web.sh
flying-sausages:
```suggestion
```

scripts/install/calibre-web.sh
liaralabs: If that's the case, is it required to install pyenv on these distros?

flying-sausages: I honestly wouldn't know. what do you advise? Just install python3 on all releases and make the venv? I'll do that for now, don't see anything wrong with it.

liaralabs: I would just use `python 3 -m venv` to create the venv for all distros. Unless the dev states a minimum Python 3 version, installing pyenv just takes extra time when the system packages are good enough. 3.x doesn't really seem to suggest a minimum version

liaralabs: I would just use python 3 -m venv to create the venv for all distros. Unless the dev states a minimum Python 3 version, installing pyenv just takes extra time when the system packages are good enough. 3.x doesn't really seem to suggest a minimum version

flying-sausages:
Currently running into segfaults, more here https://github.com/janeczku/calibre-web/issues/1755

scripts/install/calibre-web.sh
flying-sausages: Got rid of the `pyenv`, going straight for just `python3`

flying-sausages:
I'm facing an issue where for some reason, there are instances where the `admin` user does not exist in calibre-web after install so until i figure out why this should not be merged

liaralabs:
This PR needs to be rebased to ensure it merges smoothly into develop. It looks like it can be automatically merged if the HEAD is put rolled back to 3c82233d316c9721128e78fe23d903bab3257cd9 (ie there have been no commits since the merge), then `rebase` onto current develop and force push rewrite history

scripts/install/calibrecs.sh
```diff
@@ -0,0 +1,120 @@
...
+    if ! CALIBRE_LIBRARY_PATH="$(swizdb get calibre/library_path)"; then
+        CALIBRE_LIBRARY_PATH="/home/$CALIBRE_LIBRARY_USER/Calibre Library"
+        swizdb set "calibre/library_path" "$CALIBRE_LIBRARY_PATH"
+    fi
+fi
+
+clbServerPath="/home/$CALIBRE_LIBRARY_USER/.config/calibrecs"
+
+_adduser() {
+    ## TODO handle programatically
```
liaralabs, line 28: What is the status of this todo?

sources/functions/calibre
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+## Unused, left for another day
```
liaralabs, line 3: What is the status of this entire file? I would prefer not to have whole swaths of functions which are pulled in but not used anywhere.

liaralabs (changes requested):

sources/functions/calibre
flying-sausages: Unfortunately I was no able to compile calibre from source for arm :( the apt versions are old af. Unless you know how we could do it, let's drop it and see if someone comes asking for it.

scripts/install/calibrecs.sh
flying-sausages: We can use this in case we're able to get the compilation for arm going, because the dev added glags for us to do the userdb commands programmatically. the old versions in apt don't support it. Therefore until we have a thing for this, the server is protected under basic auth.

scripts/install/calibrecs.sh
```diff
@@ -0,0 +1,98 @@
...
+    cat > /etc/systemd/system/calibrecs.service << CALICS
+[Unit]
+Description=calibre content server
+After=network.target
+
+[Service]
+Type=simple
+User=$CALIBRE_LIBRARY_USER
+Group=$CALIBRE_LIBRARY_USER
+ExecStart=/usr/bin/calibre-server --max-opds-items=30 --max-opds-ungrouped-items=100 --port 8089 "${CALIBRE_LIBRARY_PATH:=CALIBRE_LIBRARY_PATH_GOES_HERE}"
```
liaralabs, line 52:
```
"${CALIBRE_LIBRARY_PATH:=CALIBRE_LIBRARY_PATH_GOES_HERE}"
```
🤨

scripts/install/calibrecs.sh
```diff
@@ -0,0 +1,98 @@
...
+    fi
+
+}
+
+# adduser function is disabled as currently not all arches support this.
+# When arm64 is built from source or it is handled in the installer, this should be revisited
+if [[ -n $1 ]]; then
+    # users=("$1")
+    # _adduser
+    exit 0
```
liaralabs, line 84: This is technically an `exit 1` then

scripts/upgrade/calibre.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+if [ ! -f /install/.calibre.lock ]; then
+    echo_error "Calibre is not installer"
```
liaralabs, line 4:
```suggestion
    echo_error "Calibre is not installed"
```

scripts/remove/calibrecs.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+#shellcheck source=sources/functions/utils
+. /etc/swizzin/sources/functions/utils
+
+if [ -z "$CALIBRE_LIBRARY_USER" ]; then
+    if ! CALIBRE_LIBRARY_USER="$(swizdb get calibre/library_user)"; then
+        CALIBRE_LIBRARY_USER=$(_get_master_username)
+        swizdb set "calibre/library_user" "$CALIBRE_LIBRARY_USER"
```
liaralabs, line 9: I'm having trouble trying to understand why we're setting this variable in swizdb while we are removing the app.

If a user removed calibre first then this would leave your install with swizdb cruft.

scripts/install/calibreweb.sh
```diff
@@ -0,0 +1,165 @@
...
+#calibreweb installer
+
+calibrewebdir="/opt/calibreweb"
+clbWebUser="calibreweb" # or make this master user?
+
+#shellcheck source=sources/functions/utils
+. /etc/swizzin/sources/functions/utils
+
+if [[ -z $CALIBRE_LIBRARY_USER ]]; then
+    CALIBRE_LIBRARY_USER="$(_get_master_username)"
```
liaralabs, line 12: We need to query swizdb for this info

scripts/install/calibreweb.sh
```diff
@@ -0,0 +1,165 @@
...
+fi
+
+if [ -z "$CALIBRE_LIBRARY_PATH" ]; then
+    # If not specifically declared, try the default calibre lib in swizzin
```
liaralabs, line 16: This should be declared in swizdb if it's not delcared by env variables

scripts/install/calibrecs.sh
```diff
@@ -0,0 +1,98 @@
...
+    exit 1
+fi
+
+#shellcheck source=sources/functions/utils
+. /etc/swizzin/sources/functions/utils
+
+if [ -z "$CALIBRE_LIBRARY_USER" ]; then
+    if ! CALIBRE_LIBRARY_USER="$(swizdb get calibre/library_user)"; then
+        CALIBRE_LIBRARY_USER=$(_get_master_username)
+        swizdb set "calibre/library_user" "$CALIBRE_LIBRARY_USER"
```
liaralabs, line 14: See note in calibre

liaralabs (changes requested):
It's almost there and there are some minor notes, but as a whole, it seems like the attention to `swizdb` needs to be a little more thorough.

The installer essentially is assuming that the user will only ever install all three apps together the one time if they set the `CALIBRE_LIBRARY_USER` or a custom library path. If you install calibre first and then anything else with the custom variables, your initial customizations will be ignored in the future. If you set these variables in the swizdb (which btw, you don't seem to be for custom variables) then that should be respected above the default of `master`

Furthermore, the remove needs to be specific about when variables should be cleared from swizdb and be careful about not leaving behind cruft accidentally. Perhaps even checking for other calibre related locks to see if any other calibre apps are still installed would be prudent.

scripts/install/calibrecs.sh
flying-sausages: That's a default value that I tell the user in echos as well as docs to change manually in case the library variable could not be resolved

scripts/install/calibrecs.sh
flying-sausages: I would say no as this should stay in line with the other installers, and I don't know if an error would break a `box adduser`. Doing nothing here is not a problem, as no issue occurred.

scripts/remove/calibrecs.sh
flying-sausages: > I'm having trouble trying to understand why we're setting this variable in swizdb while we are removing the app.

Yeah me too lol

scripts/install/calibre.sh
```diff
@@ -0,0 +1,69 @@
...
+    if ! CALIBRE_LIBRARY_USER="$(swizdb get calibre/library_user)"; then
+        CALIBRE_LIBRARY_USER=$(_get_master_username)
+        swizdb set "calibre/library_user" "$CALIBRE_LIBRARY_USER"
+    fi
+fi
+
+if [ -z "$CALIBRE_LIBRARY_PATH" ]; then
+    if ! CALIBRE_LIBRARY_PATH="$(swizdb get calibre/library_path)"; then
+        CALIBRE_LIBRARY_PATH="/home/$CALIBRE_LIBRARY_USER/Calibre Library"
+        swizdb set "calibre/library_path" "$CALIBRE_LIBRARY_PATH"
```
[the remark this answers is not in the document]
flying-sausages: Yeah good catch, should be fixed now

PR #404, opened 2020-09-17
Apt function overhaul

liaralabs:


sources/functions/apt
```diff
@@ -1,6 +1,300 @@
...
+            -*) echo "Error: Invalid option: $1"
+                exit 1
+                ;;
+            *) _apt_packages+=("$1")
+                ;;
+        esac
+        shift
+    done
+
+    if [[ -z ${_apt_packages[1]} ]]; then
```
flying-sausages, line 56: this needs to be `0` and not `1`

flying-sausages: or `${#_apt_packages} -gt 0`

sources/functions/apt
```diff
@@ -1,6 +1,300 @@
...
+        export _apt_packages
```
flying-sausages, line 57: maybe not worth exporting? Gotta test this without

PR #424, opened 2020-10-19
wip: jellyfin and jellyfin-ffmepg installer script using official repo

userdocs:


scripts/install/jellyfin.sh
```diff
@@ -46,47 +37,24 @@ create_self_ssl "${username}"
...
-# Copy the required binaries to the ffmpeg installation folder.
-cp "$install_tmp/$ffmpeg_dir_name"/* "$install_ffmpeg"
-#
-# Set the correct permissions
-chmod -R 700 "$install_ffmpeg"
-#
-# Removes the installation temporary folder as we no longer need it.
-rm -rf "$install_tmp" > /dev/null 2>&1
+apt-get update
+apt-get install jellyfin jellyfin-ffmpeg -y
```
flying-sausages, line 48:
```bash
apt_update #forces apt refresh
apt_install jellyfin jellyfin-ffmpeg
```

scripts/install/jellyfin.sh
```diff
@@ -108,219 +76,56 @@ CONFIG
...
-chown "${username}.${username}" -R "$install_dir"
-chown "${username}.${username}" -R "$install_ffmpeg"
 chown "${username}.${username}" -R "/home/${username}/.config"
 chown "${username}.${username}" -R "/home/${username}/.ssl"
+chmod -R g+rw "/home/${username}/.config/Jellyfin"
+chmod -R g+r "/home/${username}/.ssl"
+#
+# COnfigure the jellyfin service.
+#
+systemctl stop "jellyfin.service" >> /dev/null 2>&1
```
flying-sausages, line 110: `systemctl -q stop jellyfin` ?

scripts/install/jellyfin.sh
```diff
@@ -108,219 +76,56 @@ CONFIG
...
+#
+cat > "/etc/systemd/system/jellyfin.service.d/jellyfin.service.conf" <<-CONFIG
+# Jellyfin systemd configuration options
+
+# Use this file to override the user or environment file location.
+
+[Service]
+#User = jellyfin
```
flying-sausages, line 118: Does it run as `jellyfin` if this is commented? won't it run as root otherwise?

scripts/install/jellyfin.sh
```diff
@@ -108,219 +76,56 @@ CONFIG
...
+#
+cp -r /etc/default/jellyfin /etc/default/jellyfin-swizzin
+sed -r "s|JELLYFIN_CONFIG_DIR=(.*)|JELLYFIN_CONFIG_DIR=\"/home/${username}/.config/Jellyfin/config\"|g" -i /etc/default/jellyfin-swizzin
+#
+usermod -a -G ${username} jellyfin
 #
-# Enable and start the jellyfin service.
 systemctl daemon-reload
-systemctl enable --now "jellyfin.service" >> /dev/null 2>&1
+systemctl start "jellyfin.service" >> /dev/null 2>&1
```
flying-sausages, line 128: Should this not be `enable --now` in case this was a new install?

scripts/install/jellyfin.sh
```diff
@@ -108,219 +76,56 @@ CONFIG
...
     systemctl reload nginx
 fi
 #
 # Set the correct and required permissions of any directories we created or modified.
-chown "${username}.${username}" -R "$install_dir"
-chown "${username}.${username}" -R "$install_ffmpeg"
 chown "${username}.${username}" -R "/home/${username}/.config"
 chown "${username}.${username}" -R "/home/${username}/.ssl"
+chmod -R g+rw "/home/${username}/.config/Jellyfin"
+chmod -R g+r "/home/${username}/.ssl"
```
flying-sausages, line 106: where is this dir made?

flying-sausages (changes requested):

flying-sausages:
for `$username` you can use `_get_master_user` from the `utils` function, and for the password use `_get_user_password ${username}`

scripts/install/jellyfin.sh
```diff
@@ -19,15 +19,6 @@ app_port_https="$(shuf -i 10001-32001 -n 1)" && while [[ "$(ss -ln | grep -co ''
 # Get our exertnal IP 4 address and set it as a variable.
 ip_address="$(curl -s4 icanhazip.com)"
```
flying-sausages, line 20: Is this necessary? I'd skip the echos at the end altogether

flying-sausages (changes requested):

flying-sausages:
It might be nice to look at a function which would generate random ports given a range, and check they are available so that we can re-use and fool-proof the logic efficiently

scripts/install/jellyfin.sh
userdocs: This is the default ENV after it;s copied and modified.

So assume yes as i'm not trying to override anything in the area.

```
# Jellyfin default configuration options
# This is a POSIX shell fragment

# Use this file to override the default configurations; add additional
# options with JELLYFIN_ADD_OPTS.

# Under systemd, use
#  /etc/systemd/system/jellyfin.service.d/jellyfin.service.conf
# to override the user or this config file's location.

#
# General options
#

# Program directories
JELLYFIN_DATA_DIR="/var/lib/jellyfin"
JELLYFIN_CONFIG_DIR="/home/username/.config/Jellyfin/config" < will be unique to the server
JELLYFIN_LOG_DIR="/var/log/jellyfin"
JELLYFIN_CACHE_DIR="/var/cache/jellyfin"

# web client path, installed by the jellyfin-web package
JELLYFIN_WEB_OPT="--webdir=/usr/share/jellyfin/web"

# Restart script for in-app server control
JELLYFIN_RESTART_OPT="--restartpath=/usr/lib/jellyfin/restart.sh"

# ffmpeg binary paths, overriding the system values
JELLYFIN_FFMPEG_OPT="--ffmpeg=/usr/lib/jellyfin-ffmpeg/ffmpeg"

# [OPTIONAL] run Jellyfin as a headless service
#JELLYFIN_SERVICE_OPT="--service"

# [OPTIONAL] run Jellyfin without the web app
#JELLYFIN_NOWEBAPP_OPT="--nowebclient"

#
# SysV init/Upstart options
#

# Application username
JELLYFIN_USER="jellyfin"
# Full application command
JELLYFIN_ARGS="$JELLYFIN_WEB_OPT $JELLYFIN_RESTART_OPT $JELLYFIN_FFMPEG_OPT $JELLYFIN_SERVICE_OPT $JELLYFIN_NOWEBAPP_OPT"
```

Specifically `JELLYFIN_USER="jellyfin"`

scripts/install/jellyfin.sh
userdocs: I am not installing or starting the service. apt is. It's already enabled.

scripts/install/jellyfin.sh
flying-sausages: I think those do not affect the `systemd` though, and I don't see the variable used in the `systemd` file either

scripts/install/jellyfin.sh
userdocs: 1: https://github.com/liaralabs/swizzin/pull/424/files#diff-1dd9197f2356f16752770c87f5fca467a14da6824ef3da7d92515209de38f17fR41

2: by the ssl function

scripts/install/jellyfin.sh
flying-sausages: oooh gotcha.

scripts/install/jellyfin.sh
userdocs: Here is the default service.

```
[Unit]
Description = Jellyfin Media Server
After = network.target

[Service]
Type = simple
EnvironmentFile = /etc/default/jellyfin
User = jellyfin
ExecStart = /usr/bin/jellyfin ${JELLYFIN_WEB_OPT} ${JELLYFIN_RESTART_OPT} ${JELLYFIN_FFMPEG_OPT} ${JELLYFIN_SERVICE_OPT} ${JELLYFIN_NOWEBAPP_OPT}
Restart = on-failure
TimeoutSec = 15

[Install]
WantedBy = multi-user.target
```

flying-sausages: Oh I just noticed this is an override, not the actual systemd file

scripts/install/jellyfin.sh
userdocs: How do i know which port to access the application on?

With nginx i can guess the url to be /jellyfin

Without nginx i can't guess the random port.

scripts/install/jellyfin.sh
flying-sausages: /resolved

scripts/install/jellyfin.sh
flying-sausages: /resolved

scripts/install/jellyfin.sh
flying-sausages: /resolved

scripts/install/jellyfin.sh
flying-sausages: Only thing I see it in are the echos at the bottom, which I think are not necessary, barely any other apps announce where they are accessible with an actual IP. It could also check for the domain, it could as well be inaccessible outside the local network.

scripts/install/jellyfin.sh
userdocs: done.

scripts/install/jellyfin.sh
userdocs: done

scripts/install/jellyfin.sh
flying-sausages: /resolved

scripts/remove/jellyfin.sh
```diff
@@ -1,29 +1,34 @@
...
-remove_jellyfin
+#!/bin/bash
+. /etc/swizzin/sources/functions/utils
+#
+log="/root/logs/swizzin.log"
+username="$(_get_master_username)"
+#
+systemctl -q stop "jellyfin.service"
+#
+apt_remove --purge jellyfin jellyfin-ffmpeg
```
flying-sausages, line 9: without an update script, this will fail every time and `exit 1` when the package is not installed via apt as `apt_remove` kills parents script when one of the checks fails

flying-sausages:
I think we should not let users choose whether or not to upgrade.

We would have to support both of these scenarios. It would be cleaner and simpler to involuntarily move everyone to the apt-based install the next time they run `box update`, and then we can keep assuming that everyone is on the apt-based JF scenario. After all, the user won't feel a difference when they're switched over from what you've mentioned.

userdocs:
They are not being offered a choice nor did I plan to provide it.

Currently the only problem is the jellyfin data dir and files. I can show  you in detail when I push the fixed upgrade file tomorrow

flying-sausages:
> They are not being offered a choice nor did I plan to provide it.

The choice is being derived from whether the script is in the `upgrade` or `update` folder. All `update` scripts run when `box update` or `box upgrade blah` are ran. Scripts in the `upgrade` folder are only ran when the user runs `box upgrade scriptname`. Therefore the `upgrade` script will only run if the user runs it.

scripts/remove/jellyfin.sh
```diff
@@ -1,38 +1,20 @@
...
-    [[ -d "/var/log/jellyfin" ]] && rm -rf /var/log/jellyfin | :
-    [[ -d "/var/cache/jellyfin" ]] && rm -rf /var/cache/jellyfin | :
-    [[ -d "/usr/share/jellyfin/web" ]] && rm -rf /usr/share/jellyfin/web | :
-fi
+apt_remove --purge jellyfin jellyfin-ffmpeg
+#
+[[ -d "/var/lib/jellyfin" ]] && rm -rf "/var/lib/jellyfin" || :
+[[ -d "/var/log/jellyfin" ]] && rm -rf "/var/log/jellyfin" || :
+[[ -d "/var/cache/jellyfin" ]] && rm -rf "/var/cache/jellyfin" || :
+[[ -d "/usr/share/jellyfin/web" ]] && rm -rf "/usr/share/jellyfin/web" || :
```
flying-sausages, line 10: there is a function for this ;) check `utils` for `rm_ifexists` if I recall correctly

flying-sausages: also wouldn't the `|| :` be redundant here?

scripts/remove/jellyfin.sh
```diff
@@ -1,38 +1,20 @@
...
 #
 # Remove the nginx conf and reload nginx.
-if [[ -f /install/.nginx.lock ]]; then 
-    rm -f /etc/nginx/apps/jellyfin.conf
-    systemctl -q reload nginx
+if [[ -f "/install/.nginx.lock" ]]; then
```
flying-sausages, line 13: don't think quotes are necessary here, this will never expand

scripts/update/jellyfin.sh
```diff
@@ -1,87 +1,87 @@
...
+        sed -r 's#<RequireHttps>false</RequireHttps>#<RequireHttps>true</RequireHttps>#g' -i "/etc/jellyfin/system.xml"
+        #
+        if [[ -f /etc/nginx/apps/jellyfin.conf ]]; then
+            sed -r 's#proxy_pass https://127.0.0.1:(.*);#proxy_pass https://127.0.0.1:8920;#g' -i "/etc/nginx/apps/jellyfin.conf"
+            systemctl -q restart "nginx.service"
+        fi
+        #
+        [[ -d "/home/${username}/.config/Jellyfin" ]] && rm -rf "/home/${username}/.config/Jellyfin" | :
+        [[ -d "/home/${username}/.cache/jellyfin" ]] && rm -rf "/home/${username}/.cache/jellyfin" | :
+        [[ -d "/home/${username}/.aspnet" ]] && rm -rf "/home/${username}/.aspnet" | :
```
flying-sausages, line 49: `rm_ifexists` in `utils`

scripts/update/jellyfin.sh
```diff
@@ -1,87 +1,87 @@
...
+        [[ -d "/opt/jellyfin" ]] && rm -rf "/opt/jellyfin" | :
+        [[ -d "/opt/ffmpeg" ]] && rm -rf "/opt/ffmpeg" | :
+    fi
+    #
+    # data migration to default locations.
+    if [[ -d /home/${username}/.config/Jellyfin/config ]]; then
+        mkdir -p /etc/jellyfin
+        mkdir -p /var/lib/jellyfin/{data,root,metadata}
+        #
+        [[ -d "/home/${username}/.config/Jellyfin/config" ]] && cp -fRT "/home/${username}/.config/Jellyfin/config" "/etc/jellyfin" | :
```
flying-sausages, line 30: Might be worth making a function of this for utils, something like `cp_ifexists` maybe

scripts/update/jellyfin.sh
```diff
@@ -1,87 +1,87 @@
...
+    # Set the correct and required permissions of any directories we created or modified.
+    [[ -d "/home/${username}/.ssl" ]] && chown "${username}.${username}" -R "/home/${username}/.ssl" || :
+    [[ -d "/home/${username}/.ssl" ]] && chmod -R g+r "/home/${username}/.ssl" || :
+    #
+    # Set the default permissions after we have migrated our data
+    [[ -d "/etc/jellyfin" ]] && chown -R jellyfin:jellyfin "/etc/jellyfin" || :
+    [[ -f "/etc/jellyfin/logging.json" ]] && chown jellyfin:root "/etc/jellyfin/logging.json" || :
+    [[ -d "/etc/jellyfin" ]] && chown jellyfin:adm "/etc/jellyfin" || :
+    [[ -d "/var/lib/jellyfin" ]] && chown -R jellyfin:jellyfin "/var/lib/jellyfin" || :
+    [[ -d "/var/lib/jellyfin" ]] && chown jellyfin:adm "/var/lib/jellyfin" || :
```
flying-sausages, line 77: Why are all these in the conditions? Shouldn't they be guaranteed to be installed if the `apt_install` was ran? if `apt_install` fails, the script will be killed. Might be good to move the `apt_install` as high as possible so people don't possibly end up in some in-between state

scripts/update/jellyfin.sh
```diff
@@ -1,87 +1,87 @@
...
+    #
+    # Reload systemd and start the service.
+    systemctl -q daemon-reload
+    systemctl -q start "jellyfin.service"
+    #
+    echo -e "\nJellyfin upgrade completed and service restarted\n"
+    #
+    exit
```
flying-sausages, line 85: This will kill the `box update`

flying-sausages (changes requested):

scripts/update/jellyfin.sh
flying-sausages: also I don't think I have an `adm` on my box, is this a debian thing?

scripts/remove/jellyfin.sh
userdocs: it's just out of habit to quote paths in general and conditionals in square brackets. it doesn't hurt either way.

scripts/update/jellyfin.sh
```diff
@@ -1,87 +1,87 @@
...
+    # source the functions we need for this script.
+    . /etc/swizzin/sources/functions/utils
+    #
+    # Get our main user credentials using a util function.
+    username="$(_get_master_username)"
+    #
+    # disable the service to make usre the mosrt up to date and correct service file is installed via apt.
+    systemctl -q stop "jellyfin.service"
+    #
+    ## legacy start
```
flying-sausages, line 14: You might want to wrap the entire thing here in one condition, because this script will be executed every single time `box update` will be executed. So if anyone has jellyfin lockfile present, the entire thing will get executed

flying-sausages (changes requested):

scripts/remove/jellyfin.sh
userdocs: `rm_if_exists` is there. i can use that.

In a sense yes, except it wouldn't work in a function. So i do C but the `:` does nothing. 

Really it should be:

```
if [[ -d "/var/lib/jellyfin" ]]; then
	rm -rf "/var/lib/jellyfin"
fi

File: scripts/update/jellyfin.sh [hunk 15]
```diff
@@ -9,79 +9,79 @@ if [[ -f /install/.jellyin.lock ]]; then
     username="$(_get_master_username)"
     #
     # disable the service to make usre the mosrt up to date and correct service file is installed via apt.
-    systemctl -q stop "jellyfin.service"
+    systemctl -q stop jellyfin.service
```
flying-sausages, line 12: this will still stop JF every time `box update` is ran, regardless of whether or not it has been migrated

scripts/update/jellyfin.sh
```diff
@@ -9,79 +9,79 @@ if [[ -f /install/.jellyin.lock ]]; then
...
-    #
-    # Reload systemd and start the service.
-    systemctl -q daemon-reload
-    systemctl -q start "jellyfin.service"
-    #
-    echo -e "\nJellyfin upgrade completed and service restarted\n"
-    #
-    exit
-
+    if ! check_installed jellyfin; then
```
flying-sausages, line 55: Why not use this one single condition for the entire migration?

flying-sausages (changes requested):

scripts/update/jellyfin.sh
flying-sausages: instead of the separate 3 conditions which each does its own thing

scripts/remove/jellyfin.sh
flying-sausages: /resolved

scripts/update/jellyfin.sh
flying-sausages: /resolved because fuck users

scripts/update/jellyfin.sh
flying-sausages: /resolved fuck github

scripts/remove/jellyfin.sh
flying-sausages: /resolved

scripts/update/jellyfin.sh
```diff
@@ -0,0 +1,75 @@
...
+        sed -r 's#proxy_pass https://127.0.0.1:(.*);#proxy_pass https://127.0.0.1:8920;#g' -i /etc/nginx/apps/jellyfin.conf
+        systemctl -q restart nginx.service
+    fi
+    #
+    rm_if_exists "/home/${username}/.config/Jellyfin"
+    rm_if_exists "/home/${username}/.cache/jellyfin"
+    rm_if_exists "/home/${username}/.aspnet"
+fi
+#
+if ! check_installed jellyfin; then
```
flying-sausages, line 44: Without the lockfile condition at the top, this will just install jellyfin for everyone if they run `box update`

flying-sausages (changes requested):

scripts/update/jellyfin.sh
userdocs: the package does this. i tested multiple times

adm: Group adm is used for system monitoring tasks. Members of this group can read many log files in /var/log, and can use xconsole. Historically, /var/log was /usr/adm (and later /var/adm), thus the name of the group.

flying-sausages: > the package does this. i tested multiple times

Is that a reply to why the tests aren't needed?

scripts/update/jellyfin.sh
userdocs: done

scripts/update/jellyfin.sh
flying-sausages: /resolved

scripts/update/jellyfin.sh
flying-sausages: /resolved

scripts/update/jellyfin.sh
userdocs: I removed all the `|| :` tests

flying-sausages: I was referring to the `[[ -d ... ]]` checks. I think all of these should exist, and if not then there's a problem that should be thrown at the user

scripts/update/jellyfin.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+#
+if [[ -f /install/.jellyin.lock ]]; then
```
flying-sausages, line 3: `jellyin` ;) might wanna check that's not the case through the whole thing

flying-sausages (changes requested):

scripts/update/jellyfin.sh
userdocs: done, removed.

scripts/update/jellyfin.sh
userdocs: resolved.

scripts/update/jellyfin.sh
flying-sausages: /resolved

scripts/update/jellyfin.sh
flying-sausages: /resolved

flying-sausages (approved):
my name is flyin g. sausage and I approve this PR

PR #430, opened 2020-10-26
Update readme link to:

userdocs:
https://swizzin.ltd/getting-started

some cosmetic tweaks.

Where does the donate link go? https://swizzin.ltd/#donate

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,12 @@
 
 # 2.6.0 Stable
 
-[website](https://swizzin.ltd) | [docs](https://swizzin.ltd/docs) | [discord](https://discord.gg/bDFqAUF)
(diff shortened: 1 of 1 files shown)


flying-sausages:
I intend to make the `/docs/` page soon, but the backtick lang is still a nice change. might just make a pr late rhwn i get to it to prevent 404s in the meantime

liaralabs:
We could also use the new shorter setup link in the quick start: https://git.io/swizzin-setup

liaralabs:
I will need to revisit the donation link since:
processing credit cards is against github pages ToS
It's incredibly difficult to do CC processing on static pages (perhaps stripe checkout)

However, I'm mostly just pushing github sponsors these days and deprecating the method of charging one-off donations.

PR #434, opened 2020-10-27
jellyfin and jellyfin-ffmpeg

userdocs:
Use apt-get to install the jellyfin packages from the officially maintained repositories and migrate existing installations to this setup with no loss of user data.


scripts/install/jellyfin.sh
```diff
@@ -1,33 +1,20 @@
...
-install_ffmpeg="/opt/ffmpeg"
+# Source the global functions we require for this script.
+. /etc/swizzin/sources/functions/utils
+. /etc/swizzin/sources/functions/ssl
 #
-# Set an installation temporay directory.
-install_tmp="/tmp/jellyfin"
+# Get our some useful information from functions in the sourced utils script
+username="$(_get_master_username)"
+dist_info # get our distribution ID, set to DIST_ID, and VERSION_CODENAME, set to DIST_CODENAME, from /etc/os-release
```
flying-sausages, line 17: I'd do this in a separate PR where it's applied anywhere necessary otherwise this won't be implemented right I think. Also might need a separate focus and review. I'm not the one to talk with my 150+ file PRs hahaha so let me know if you think it makes sense

scripts/update/jellyfin.sh
```diff
@@ -0,0 +1,80 @@
...
+        chown jellyfin:root "/etc/jellyfin/logging.json"
+        chown jellyfin:adm "/etc/jellyfin"
+        chown -R jellyfin:jellyfin "/var/lib/jellyfin"
+        chown jellyfin:adm "/var/lib/jellyfin"
+        #
+        # Reload systemd and start the service.
+        systemctl -q daemon-reload
+        systemctl -q start jellyfin.service
+        #
+        echo_success "The Jellyfin upgrade has completed"
```
flying-sausages, line 78: Should be "Jellyfin installed" to fit the pattern of the others

flying-sausages: And in this event I'd even use "Jellyfin updated" actually

scripts/install/jellyfin.sh
```diff
@@ -116,234 +51,51 @@ cat > "/home/${username}/.config/Jellyfin/config/dlna.xml" <<-CONFIG
...
-echo_success "Jellyfin installed"
+touch /install/.jellyfin.lock
 #
-if [[ ! -f /install/.nginx.lock ]]; then
-    echo_info "Jellyfin is available at: https://$(curl -s4 icanhazip.com):${app_port_https}"
-else
-    echo_info "Jellyfin is now available in the panel"
-    echo_info "Please visit https://$ip_address/jellyfin"
-fi
+echo_success "The Jellyfin installation has completed"
```
flying-sausages, line 99: Should be "Jellyfin installed" to be in line with the rest

scripts/update/jellyfin.sh
```diff
@@ -0,0 +1,80 @@
...
+        fi
+        #
+        rm_if_exists "/home/${username}/.config/Jellyfin"
+        rm_if_exists "/home/${username}/.cache/jellyfin"
+        rm_if_exists "/home/${username}/.aspnet"
+    fi
+    #
+    if ! check_installed jellyfin; then
+        echo_info "Updating Jellyfin"
+        #
```
flying-sausages, line 48: I'm wondering to what degree is there code duplication here between the JF installer and update. Could possibly move out the common steps into a `sources/functions` file and have both of the scripts call the function there. Could make it easier to maintain down the road.

flying-sausages: Not something I think is required but if you agree I think it would be nice

flying-sausages (changes requested):

scripts/install/jellyfin.sh
userdocs: i just moved it to the script. if the change gets added, it can be removed.

scripts/update/jellyfin.sh
userdocs: Changed to `Jellyfin updated`

scripts/install/jellyfin.sh
userdocs: changed to "Jellyfin installed"

scripts/update/jellyfin.sh
userdocs: At some point it will be obsolete and you can just remove the code. I don't the benefit of it being a separate function

scripts/upgrade/jellyfin.sh
```diff
@@ -1,68 +0,0 @@
-#!/usr/bin/env bash
```
flying-sausages, line 1: could just leave an `echo "Please use apt to upgrade jellyfin"` but that's cosmetics only

flying-sausages (approved):

scripts/upgrade/jellyfin.sh
userdocs: What is this referencing?

flying-sausages: The file in its entirety.

userdocs: The only part that will update is this, the rest is migration.

https://github.com/userdocs/swizzin/blob/master/scripts/update/jellyfin.sh#L46#L79

Does this mean the apt-get will be carried out twice. once by a function and again by this script?

flying-sausages: nono I meant that if a user attempts to run `box upgrade jellyfin` they're told to use apt instead and nothing else

userdocs: i feel like that is an extra step. as a user i'd prefer if the command gave me the result i expected. to update/fix jellyfin. we are already using build in functions to do the apt-get

scripts/update/jellyfin.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env bash
+#
+if [[ -f /install/.jellyfin.lock ]]; then
+    # source the functions we need for this script.
+    . /etc/swizzin/sources/functions/utils
+    # Get our main user credentials using a util function.
+    username="$(_get_master_username)"
+    dist_info # get our distribution ID, set to DIST_ID, and VERSION_CODENAME, set to DIST_CODENAME, from /etc/os-release
```
flying-sausages, line 8: `scripts/update/jellyfin.sh: line 8: dist_info: command not found`
Need to make sure this is sourced in

flying-sausages (changes requested):

flying-sausages (approved):
Good to merge as is

PR #439, opened 2020-10-27
Update CONTRIBUTING.md

userdocs:
echo_progress_start and /echo_progress_start/echo_progress_done/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,32 +1,42 @@
 # Contributing guidelines
+
 Here are a couple things to take into account when contributing to swizzin.
 
 ## Directories and files
+
 - Please use `/opt` to install any new software. **Avoid using home directories to install application binaries/source**
 - When creating configuration for users, please add it to an appropriate folder under `~/.config/` when possible
 - Please make sure to put any necessary code output into the default log files for swizzin, (i.e. `swizzin.log` and `install.log` under `/root/logs`) *including* the `stderr`
 - Make sure to create `.lock` files for your application under `/install/`
   - Record any application information that might need to be dynamically retrieved into the .lock file above.
 
-## Bash code styleguide
+## Bash code style guide
+
 ### Shellcheck
+
 Please use `shellcheck` and its IDE plugins and resolve any warnings for the code you have contributed.
 
 ### Code re-use
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -117,26 +137,29 @@ There is a wide choice of "styles" to choose from, please use them appropriately
   * Any generated unique values that need to match some specification (e.g. ports)
   * Any variables that are getting assigned from either the user or from the return of a command/function
 
+## Logging - further reading
```
flying-sausages, line 140: Further reading? Is there previous reading? There's a reference to Terminal output sanitation but not to logging as a whole

flying-sausages (approved):

CONTRIBUTING.md
liaralabs: There's the apt logging section above which mentions some things, but as far as a headline item would go, I'm not sure it's 100% appropriate.

Do we need to differentiate between the two?

userdocs: ## Why are we here?

This result of [linting against common mark](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint), upon moving to remark linting to match Codacy, it does not appears to be as fussy out duplicate headings but it brings to mind 2 general issues.

-   Codacy uses [remark for linting](https://github.com/codacy/codacy-remark-lint) as i found out after this pull was made. So the title change in and of itself is not relevant to this context. What matter is that

-   Why does it matter either way, when really the document needs to be rewritten to be useful to the reader, as i found out when trying to use it. It also seems like the original comment didn't consider the text that followed the title change to understand the context. Is this a comment for the sake or commenting? a late night commenting frenzy or a contextually aware criticism with focus towards making a doc aimed at ends user more helpful?

### Solution?

So the solution is to go by the outcome of something like this use <https://marketplace.visualstudio.com/items?itemName=mrmlnc.vscode-remark> which does not complain about having a duplicate title.

Create a goal towards these outcomes?

-   general improvement of the docs with better example and wording towards being easier to implement.

-   add markdown formatting guidelines?

### Further reading

I also found that this formatted with the same outcome but appears to use common mark. <https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one#github-flavored-markdown>

Perhaps the first linter was overly cautious or just wrong to complain about a separately nested title.

PR #446, opened 2020-11-14
HOTFIX: remove dumb mango file

flying-sausages:
The file broke the output of the `_get_user_list` function. We really need to restructure that whole mechanism


userdocs (approved):
fixes bad expansion of `${u}` on this line
https://github.com/swizzin/swizzin/blob/master/scripts/nginx/rutorrent.sh#L241

PR #452, opened 2020-11-15
Update settings.json

userdocs:
We can define some settings if using https://marketplace.visualstudio.com/items?itemName=foxundermoon.shell-format

Tabs at 4 spaces.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,7 +5,7 @@
         "shellcheck",
         "swizzin",
         "xenial"
-    ]
-    "shellformat.flag": "-i 4",
-    "shellformat.path": ""
+    ],
+    "shellformat.flag": "-ci -sr -i 0",
+    "editor.formatOnSave": true
 }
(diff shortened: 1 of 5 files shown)


.vscode/settings.json
```diff
@@ -6,4 +6,6 @@
         "swizzin",
         "xenial"
     ]
-}
\ No newline at end of file
+    "shellformat.flag": "-i 4",
+    "shellformat.path": ""
```
flying-sausages, line 10: This should be left out as that's relative for everyone. Leaving this unset will use plugin default, which either downloads the binary or uses it if it is in the path

userdocs: we are just talking about line 10? ` "shellformat.path": ""`

liaralabs:
> Tabs at 4 spaces.

Shots fired.

flying-sausages (approved):
I have nothing else to add to this

PR #461, opened 2020-11-16
Making utils use uniform

flying-sausages:
Ensure all use-cases covered by the functions in utilities are replaced by the correct usage

**NEEDS TO BE PROPERLY TESTED FIRST**


flying-sausages:
alright i better just do this from scratch

PR #463, opened 2020-11-16
Update jellyfin.sh

userdocs:
fix cache and metadata paths if set in the system.xml


flying-sausages (approved):
Approving on the condition this resolves the linked issue

PR #464, opened 2020-11-16
Start failing scripts if pip reqs don't install right

flying-sausages:
I hope we won't suffer from this https://github.com/pypa/pip/issues/8001

### DIFF
--- a/scripts/install/bazarr.sh
+++ b/scripts/install/bazarr.sh
@@ -9,23 +9,21 @@
 #   including (via compiler) GPL-licensed code must also be made available
 #   under the GPL along with build & install instructions.
 
-
-
 codename=$(lsb_release -cs)
 
-user=$(cut -d: -f1 < /root/.master.info )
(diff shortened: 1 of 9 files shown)


liaralabs:
We should probably be adding some extra info here:

```
"Pip failed to install the depends, please check log"
```
Or something along those lines. Since pip output is redirect and this is the command that's failing, it should be pretty clear in the logs what's going wrong.

Also, is simply quitting the script without cleaning up the way to go here? Most checks which just quit the script are run before changes are made, but I think we should be cleaning up after ourselves if we're going to start killing scripts due to failure. This could be as simple as running the uninstall script before exit. The lockfile is only required to remove the app if calling it through box -- if we call the uninstall script directly, it should be fine.

flying-sausages:
> Or something along those lines

The default `echo_error` already gives a pasteable command to check the logs. I'm gonna get back to the cleanup behavior in a sec after checking the files again

flying-sausages:
This is probably unnecessary

PR #465, opened 2020-11-17
Update key setting

Thomvh:
Update the way to set the key to the x2go preferred method.

### DIFF
--- a/scripts/install/x2go.sh
+++ b/scripts/install/x2go.sh
@@ -42,7 +42,8 @@ deb-src http://packages.x2go.org/debian ${release} main
 #deb-src http://packages.x2go.org/debian ${release} heuler
 EOF
     echo_progress_done "Repo added"
-    apt-key --keyring /etc/apt/trusted.gpg.d/x2go.gpg adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys E1F958385BFE2B6E >>${log} 2>&1
+    #apt-key --keyring /etc/apt/trusted.gpg.d/x2go.gpg adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys E1F958385BFE2B6E >>${log} 2>&1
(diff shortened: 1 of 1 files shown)


flying-sausages:
Heya, thanks! I'm assuming you tested this, can you just confirm that? I'm assuming this will work the same on Deb an Ubu

flying-sausages:
When I run the line added on my 20.04, I get the following. Looks like it still needs a keyring?

```
Executing: /tmp/apt-key-gpghome.bBV9iczrdh/gpg.1.sh --recv-keys --keyserver keys.gnupg.net E1F958385BFE2B6E
gpg: key E1F958385BFE2B6E: public key "X2go Debian/Ubuntu Packaging <debian@x2go.org>" imported
gpg: Total number processed: 1
gpg:               imported: 1
gpg: no writable keyring found: Not found
gpg: error reading '[stdin]': General error
gpg: import from '[stdin]' failed: General error
gpg: Total number processed: 0
```

flying-sausages:
I'm wondering if it would be a good idea to check the exit code of this and throw an `echo_error` and an `exit 1` if it fails as well

liaralabs:
I don't see the need to change this command.

The proposed change is basically only changing the keyserver and storing the key in the main keystore rather than the one specified for x2go.

There were issues with the key import and apt update but I believe this was patched yesterday. The repo was being updated before the key was added which was causing apt to fail. Now, apt is not updated until the key is added to sign the x2go repo.

liaralabs:
> When I run the line added on my 20.04, I get the following. Looks like it still needs a keyring?

This command is not for Ubuntu at all. Ubuntu uses a ppa.

flying-sausages:
> This command is not for Ubuntu at all

Oh yeah you're right, should have checked the rest of the file ha

flying-sausages:
> this was patched yesterday

Let's close this PR if we don't get any reports about this failing in the next 3/4 weeks I say

flying-sausages:
Soz

PR #468, opened 2020-11-21
Fix qbittorrent regex to include minor bugfix versions

Anakros:
The qBittorrent script doesn't support specific qBittorrent versioning and doesn't pull the latest version:
```
curl -s https://github.com/qbittorrent/qBittorrent/releases | grep .tar.gz | grep -oP '\d.\d?.?\d'
4.3.0
4.3.0
4.2.5
4.2.4
4.2.3
4.2.2
4.2.1
4.2.0
4.1.9
4.1.9
```

```
curl -s https://github.com/qbittorrent/qBittorrent/releases | grep .tar.gz | grep -oP '\d.\d?.?\d?.?\d'
4.3.0.1
4.3.0
4.2.5
4.2.4
4.2.3
4.2.2
4.2.1
4.2.0
4.1.9.1
4.1.9
```

### DIFF
--- a/sources/functions/qbittorrent
+++ b/sources/functions/qbittorrent
@@ -5,11 +5,12 @@ function whiptail_qbittorrent() {
 		. /etc/swizzin/sources/functions/apt
 		codename=$(lsb_release -cs)
 		repov=$(get_candidate_version qbittorrent-nox)
-		latestv41=$(curl -s https://github.com/qbittorrent/qBittorrent/releases | grep -m1 -oP '4\.1\.\d')
-		if [[ -z $latestv41 ]]; then latestv41=4.1.9; fi
-		latestv42=$(curl -s https://github.com/qbittorrent/qBittorrent/releases | grep -m1 -oP '4\.2\.\d')
-		if [[ -z $latestv41 ]]; then latestv42=4.2.5; fi
-		latestv=$(curl -s https://github.com/qbittorrent/qBittorrent/releases | grep -m1 .tar.gz | grep -oP '\d.\d?.?\d')
+		releases=$(git ls-remote -t --refs https://github.com/qbittorrent/qBittorrent.git | awk '{sub("refs/tags/release-", ""); print $2 }' | sort -r)
+
+		latestv41=$(echo "$releases" | grep -m1 -oP '4\.1\.\d?.?\d')
+		latestv42=$(echo "$releases" | grep -m1 -oP '4\.2\.\d?.?\d')
+		latestv=$(echo "$releases" | grep -m1 -oP '\d.\d?.?\d?.?\d')
+
 		case ${codename} in
 			"xenial" | "stretch")
 				function=$(whiptail --title "Install Software" --menu "Choose a qBittorrent version:" --ok-button "Continue" --nocancel 12 50 3 \


flying-sausages:
Can this be changed to use the API instead of parsing html with grep? `jq` is available on all box installs by now.

If you want to use the tags, you can do this for example

```
➜  ~ curl -s https://api.github.com/repos/qbittorrent/qBittorrent/tags | jq -r '.[].name' | cut -d- -f2

4.3.0.1
4.3.0
4.2.5
4.2.4
4.2.3
4.2.2
4.2.1
4.2.0
4.1.9.1
4.1.9
4.1.8
4.1.7
4.1.6
4.1.5
4.1.4
4.1.3
4.1.2
4.1.1
4.1.0
4.0.4
4.0.3
4.0.2
4.0.1
4.0.0
3.3.16
3.3.15
3.3.14
3.3.13
3.3.12
3.3.11
```

Anakros:
Made changes to use GitHub's API instead as suggested.

userdocs:
So any thoughts on doing it like this?

```bash
grep -Eom 1 'release-([0-9]{1,4}\.?)+$' <(curl -sNL "https://github.com/qbittorrent/qBittorrent/tags")
release-4.3.0.1

grep -Eo 'release-([0-9]{1,4}\.?)+$' <(curl -sNL "https://github.com/qbittorrent/qBittorrent/tags")
release-4.3.0.1
release-4.3.0
release-4.2.5
release-4.2.4
release-4.2.3
release-4.2.2
release-4.2.1
release-4.2.0
release-4.1.9.1
release-4.1.9
```

flying-sausages:
I'll abstain from reviewing this one for now.

We've discussed this and we'd like to merge it, but `4.3.0.1` currently has no changes for Linux. That doesn't mean there won't be a release in the future that will of course.

Someone else might want to test and review this however.

Anakros:
I don't see how is it relevant what changes are there, because there is no official indication that the X.Y.Z.A versions are windows-only. It's clearly a bug that swizzin doesn't list all versions available no matter what changes were made in them.

userdocs:
https://www.qbittorrent.org/news.php

image

1: no urgnecy for this fix.
2: we are discussing whether to use the api or an alternative method

The api has rate limiting the web scraping does not. There are other considerations and no reason to rush.

Anakros:
There is no rush :)
I like the `git ls-remote` solution the most, btw.

userdocs:
That seems to work well for qbittorrent as they tag consistently and i have some examples.

Set all tags to an array.

```bash
qb_versions=($(git ls-remote -t --refs https://github.com/qbittorrent/qBittorrent.git | awk '{sub("refs/tags/release-", ""); print $2 }'))
```

Echo only items starting with 4.

```bash 
echo ${qb_versions[@]##[^4]*}
4.0.0 4.0.1 4.0.2 4.0.3 4.0.4 4.1.0 4.1.1 4.1.2 4.1.3 4.1.4 4.1.5 4.1.6 4.1.7 4.1.8 4.1.9 4.1.9.1 4.2.0 4.2.1 4.2.2 4.2.3 4.2.4 4.2.5 4.3.0 4.3.0.1
```

(bash 4.4+) echo last item.

```bash
echo ${qb_versions[-1]}
4.3.0.1
```

It seems like it work well for qbit.

liaralabs:
As long as we can keep the original versioning intact then I think the `git remote-ls` seems like it may be the best option at the moment and one that doesn't rely on an API which could rate limit you. 

As of now, the script should pull:
- Latest
- Latest 4.2
- Latest 4.1

Problem is, it's not quite as easy to work with an array of version strings, especially since the tags are listed oldest to newest

```
qb_versions=($(git ls-remote -t --refs https://github.com/qbittorrent/qBittorrent.git | awk '{sub("refs/tags/release-", ""); print $2 }' | sort -r))
```
Gets us the array so that the latest is position 0, but it looks like we now have to call a for loop in order to drill down to the previous versions we need, which is definitely more complicated than what's already going on.

userdocs:
Since there is a predictable consistency with the way qbitorrent tag and name their releases i feel like a expanded version of the original method, to just grab the extra .1 is acceptable for this scenario.

The alternative `git ls-remote` does seems to require an additional complexity that was not present to start with and i am not sure we are fixing anything by introducing it, in this specific example.

Anakros:
I don't think the array of strings is needed unless I miss something. Just save the `ls-remote` output as text and grep it as usual:
```bash
$ qb_versions=$(git ls-remote -t --refs https://github.com/qbittorrent/qBittorrent.git | awk '{sub("refs/tags/release-", ""); print $2 }')
$ echo "$qb_versions" | sort -r | grep -m1 -oP '4\.2\.\d?.?\d'
4.2.5
$ echo "$qb_versions" | sort -r | grep -m1 -oP '4\.3\.\d?.?\d'
4.3.0.1
$ echo "$qb_versions" | sort -r | grep -m1 -oP '4\.1\.\d?.?\d'
4.1.9.1
$ echo "$qb_versions" | sort -r | grep -m1 -oP '\d.\d?.?\d?.?\d'
4.3.0.1
```

liaralabs:
You're right, I was indeed overthinking it. Saving the versions as a string instead of an array lets us grab the info we need without getting further complicated. Bonus, we don't have to worry about pagination from `curl`ing the releases page.

Since it's sounding like we're reaching a quorum on this issue, I think we should move forward with `git ls-remote`. I think we can safely remove the following lines since we don't have to worry about pagination now as the `ls-remote` lists all tags:

```
    if [[ -z $latestv41 ]]; then latestv41=4.1.9; fi
```

```
    if [[ -z $latestv41 ]]; then latestv42=4.2.5; fi
```

Aaaand look, it has a typo in the variable anyway :sweat_smile:

Anakros:
I updated the script with the proposed solution.

userdocs:
ha, this was bad timing: https://github.com/qbittorrent/qBittorrent/releases/tag/release-4.3.1

``
4.3.1
4.3.0.1
4.3.0
4.2.5
```

But the code works, so i am ok with the change.

**Comment by userdocs** [comment]
I approve this.

```bash
~ : releases=$(git ls-remote -t --refs https://github.com/qbittorrent/qBittorrent.git | awk '{sub("refs/tags/release-", ""); print $2 }' | sort -r)
~ : latestv41=$(echo "$releases" | grep -m1 -oP '4\.1\.\d?.?\d')
~ : latestv42=$(echo "$releases" | grep -m1 -oP '4\.2\.\d?.?\d')
~ : latestv=$(echo "$releases" | grep -m1 -oP '\d.\d?.?\d?.?\d')
~ : echo $latestv41
4.1.9.1
~ : echo $latestv42
4.2.5
~ : echo $latestv
4.3.1
~ : latestv="4.3.0.1"
~ : echo $latestv
4.3.0.1
~ :
```

userdocs (approved):

flying-sausages:
@Anakros FYI we just pushed a formatting update, please run your PR through `shfmt` if you haven't yet (check the Contrbutors.md file if you haven't yet). Cheers!

Anakros:
@flying-sausages it's done.

PR #469, opened 2020-11-21
Utils - new function - dist_info

userdocs:
Get dist codename and id from `/etc/os-release`

### DIFF
--- a/sources/functions/utils
+++ b/sources/functions/utils
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+function dist_info() {
+	DIST_CODENAME="$(source /etc/os-release && echo "$VERSION_CODENAME")"
+	DIST_ID="$(source /etc/os-release && echo "$ID")"
+}
+
 function install_rar() {
 	if [[ -z $(which rar) ]]; then
 		apt_install rar unrar
--- a/sources/functions/utils
+++ b/sources/functions/utils
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 function dist_info() {
-	DIST_CODENAME="$(source /etc/os-release && echo "$VERSION_CODENAME")"
-	DIST_ID="$(source /etc/os-release && echo "$ID")"
+	OS_CODENAME="$(source /etc/os-release && echo "$VERSION_CODENAME")"
+	OS_DISTRO="$(source /etc/os-release && echo "$ID")"
 }
 
 function install_rar() {


flying-sausages:
Ooooh does that mean we can stop installing `lsb_release` on setup? Could make the installer a good bit faster. We'll have to handle the case when it's missing too though.

userdocs:
i guess, but adding this function should not change or break anything but be considered an additional option for now.

It just means i can edit remove it from the jellyfin scripts.

flying-sausages:
> should not change or break anything

I think if we can just rewrite the bit in `setup.sh` to skip the install that would already be a nice improvement on its own. I'm sure this will work. Together with a good "try-catch" layout where we cleanly exit if the function errors out for some reason" it would be a good addition.

That's the whole reason why the whole setup pauses for easily 2 minutes on slower machines/connections before anything happens at all. It's being installed just so we can tell people "You can't run this script" really.

liaralabs:
> Ooooh does that mean we can stop installing `lsb_release` on setup? Could make the installer a good bit faster. We'll have to handle the case when it's missing too though.

I don't think this is going to shave minutes off the setup :rofl: The setup is also running `apt upgrade` which can take some time if the OS image was not fully up-to-date when the server was deployed. `lsb_release` is just a tiny app, takes seconds to install

liaralabs:
So we are just adding the function for now, and then we will start pushing the new standard?

userdocs:
yeah, i can then edit the jellyfin script to remove the built in version. there should be no conflicts.

sources/functions/utils
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+function dist_info() {
+	DIST_CODENAME="$(source /etc/os-release && echo "$VERSION_CODENAME")"
```
liaralabs, line 4: Couple things here:

- I'm not 100% certain it makes sense to have this as a separate sub-function of utils. Could we not just have the variables defined if utils is imported?
- Do we need to rename variables?

userdocs: it should be moved out of utils altogether then and set at the earliest opportunity as exported variables and not a function. Where would that be?

I had just `id="$(source /etc/os-release && echo "$ID")"` and i think this conflicts with certain scripts.

Maybe it's not so great to export as the same or uppercase so.

```
os_codename="$(source /etc/os-release && echo "$VERSION_CODENAME")"
os_id="$(source /etc/os-release && echo "$ID")"
```

liaralabs: Hmm, well I've already thought that we should have a case for things like `apt` functions which are sourced globally through `box`. One function to source all our global depends (`echos`, `apt`, `dist_info`) which we can also just source in setup after the repo has been cloned locally.

I guess `ID` is maybe not the *most* unique variable. Easier to choose unique globals than worry about clobbering simple named ones in app scripts.

userdocs: Since it's a file we are sourcing and there is no automation of these loaded as system vars, i don't think renaming them to something unique will do any harm. 

Really, if you wanted that, you could just source the file itself in box.

```bash
source /etc/os-release
```

But for the 2 vars we require i can't see it being a problem as long as the naming is predictable.

Maybe it should just go in box as.

```bash
export os_codename="$(source /etc/os-release && echo "$VERSION_CODENAME")"
export os_id="$(source /etc/os-release && echo "$ID")"
```

sources/functions/utils
```diff
@@ -1,5 +1,10 @@
...
+	DIST_ID="$(source /etc/os-release && echo "$ID")"
```
flying-sausages, line 5:
```suggestion
	OS_CODENAME="$(source /etc/os-release && echo "$VERSION_CODENAME")"
	OS_DISTRO="$(source /etc/os-release && echo "$ID")"
```

I think this would give more context and clarity over what these contain

flying-sausages:
```suggestion
	DIST_CODENAME="$(grep VERSION_CODENAME /etc/os-release | cut -d= -f2)"
	DIST_ID="$(grep -E "^ID=" /etc/os-release | cut -d= -f2)"
```

If we want to make sure that we don't actually source the `ID` value. Not sure if it will or won't so a separate suggestion

flying-sausages (changes requested):
There's two things I think should be in this PR still:
- A way to make these values available underneath everything ran from `box` and `setup.sh`
    - Will we source this entire file above, trigger the function and then export them individually?
    - Should we make a different script file which just exports these variables, and then source that in the scripts?

There should be some "test" on how easily this will be accessed

sources/functions/utils
userdocs: I think sourcing it in subshell has the same outcome as grepping it in this case but maybe a bit more predictable, but worth keeping in mind though.

PR #470, opened 2020-11-22
jellyfin migration fix:

userdocs:
Make sure user's libraries are fully migrated by editing paths in the library.db. These changes are determined by the existence of library folders in the data folder.

Please see this commit for specific changes to avoid getting lost in the formatting change.

https://github.com/swizzin/swizzin/commit/76fcb0ae9a30cfdb2fd339938113972f240b4905

### DIFF
--- a/scripts/update/jellyfin.sh
+++ b/scripts/update/jellyfin.sh
@@ -18,7 +18,7 @@ if [[ -f /install/.jellyfin.lock ]]; then
         echo_progress_start "Removing old Jellyfin service"
         systemctl -q disable --now jellyfin.service
         rm_if_exists /etc/systemd/system/jellyfin.service
-        kill -9 $(ps xU ${username} | grep "/opt/jellyfin/jellyfin -d /home/${username}/.config/Jellyfin$" | awk '{print $1}') >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


scripts/update/jellyfin.sh
```diff
@@ -1,93 +1,104 @@
...
-        systemctl -q daemon-reload
-        systemctl -q start jellyfin.service
-        #
-        echo_success "Jellyfin updated"
-    fi
+	# awaiting pull to remove
+	function dist_info() {
+		DIST_CODENAME="$(source /etc/os-release && echo "$VERSION_CODENAME")"
+		DIST_ID="$(source /etc/os-release && echo "$ID")"
+	}
```
flying-sausages, line 8: think you might want to remove this from this PR if it's already in the other one

flying-sausages (changes requested):
Didn't have the chance to go through the rest of it with enough focus seeing as it's hard to see what's formatting and what isn't. Let me know if I should do a double-pass later. Or maybe we can check a diff of this file against a formatted but otherwise untouched version?

scripts/update/jellyfin.sh
userdocs: the function need to me merged first.

flying-sausages (approved):

PR #473, opened 2020-11-23
Formatting check GitHub action

flying-sausages:
Github action to check .sh formatting (and eventually shellcheck too)

### DIFF
--- a/.github/workflows/sh-checker.yml
+++ b/.github/workflows/sh-checker.yml
@@ -10,7 +10,7 @@ jobs:
         uses: luizm/action-sh-checker@v0.1.8
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          SHELLCHECK_OPTS: -e SC1004 # It is posible to exclude some shellcheck warnings.
+#           SHELLCHECK_OPTS: -e SC1004 # It is posible to exclude some shellcheck warnings.
           SHFMT_OPTS: -ci -sr -i 0 # It is posible to pass arguments to shftm
         with:
           sh_checker_comment: true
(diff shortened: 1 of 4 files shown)


flying-sausages:
Maybe I can set this up to only check the Files Changed? https://github.com/luizm/action-sh-checker/issues/27

.github/workflows/sh-checker.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  sh-checker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Run the sh-checker
+        uses: luizm/action-sh-checker@v0.1.8
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SHELLCHECK_OPTS: -e SC1004 # It is posible to exclude some shellcheck warnings.
```
userdocs, line 13: Do we need this particular exclusion? I know it's in the example action but is it required?

https://github.com/koalaman/shellcheck/wiki/SC1004

flying-sausages: Ah lol nope not at all. It should pick up shellcheckrc

liaralabs:
Another shellchecker? Can we potentially trim down since we now have 3 checkers in the repo?

flying-sausages:
> Another shellchecker? Can we potentially trim down since we now have 3 checkers in the repo?

I don't see any other checker being ran under any of the other PRs right now. This is the only one we have here. And with the current implementation, it does `shfmt` checks **only**

userdocs (approved):

PR #479, opened 2020-11-25
function os id and codename

userdocs:


sources/functions/os
```diff
@@ -0,0 +1,9 @@
+_os_id() {
+	source /etc/os-release
+	echo "$ID"
+}
```
flying-sausages, line 4:
```suggestion
_os_distro() {
	source /etc/os-release
	echo "$ID"
}
export -f _os_distro
```

sources/functions/os
```diff
@@ -0,0 +1,9 @@
...
+
+_os_codename() {
+	source /etc/os-release
+	echo "$VERSION_CODENAME"
+}
```
flying-sausages, line 9:
```suggestion
_os_codename() {
	source /etc/os-release
	echo "$VERSION_CODENAME"
}
export -f _os_codename
```

flying-sausages (commented):
In a script, you could then do

```bash
if [[ $(_os_distro) == "ubuntu" ]]; then #...
```

sources/functions/os
flying-sausages: I'm also debating `_os_version` maybe? would that be too obfuscated?

flying-sausages (commented):
I'd add comments above the functions which describe what they do and the expectable output (including when it has a newline or not maybe?) but that's a detail

flying-sausages (approved):

PR #512, opened 2020-12-10
Moviematch

flying-sausages:
## Summary
Adds moviematch app, without baseurl support (but all the other groundwork for it) as it doesn't support it yet (https://github.com/LukeChannings/moviematch/issues/10)

## Fixes issues: 
- N/A

## Proposed Changes:
- Add moviematch app
- Add installer for deno dependency

## Categories

- New feature (non-breaking change which adds functionality)

## Checklist
- [ ] Docs have been made/are not necessary
    - PR link: 
- [ ] Changes to panel have been made/are not necessary
    - PR link: 
- [ ] Code is formatted [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)

## Testing done
OS & Version: Ubuntu 20.04

### How have you tested this

Scenarios tested:
- Install
- Remove

### DIFF
--- a/scripts/install/moviematch.sh
+++ b/scripts/install/moviematch.sh
@@ -1,9 +1,16 @@
 #!/bin/bash
 
-# if [[ -f /install/,plex.lock ]]; then
-# 	echo_error "This app is useless without Plex"
-# 	exit 1
-# fi
+if [[ ! -f /install/.plex.lock ]]; then
+	echo_warn "Plex server not installed"
+	echo_query "Please enter your Plex Server's address (e.g. domain.ltd:32400)"
+	read -r plexaddress
+else
+	plexaddress="localhost:32400"
+fi
+
+echo_info "Moviematch needs to connect to Plex via a Token. You can hear how to retrieve one here:\nhttps://support.plex.tv/articles/204059436-finding-an-authentication-token-x-plex-token/"
+echo_query "Please enter your Plex Token"
+read -r plextoken
 
 #shellcheck source=sources/functions/deno
 . /etc/swizzin/sources/functions/deno
@@ -22,15 +29,10 @@ useradd moviematch --system -d "$mmatchDir" >> $log 2>&1
 sudo chown -R moviematch:moviematch $mmatchDir
 
 cat > $mmatchDir/.env << ENV
-# !!! Change the PLEX_TOKEN in this file, then close the editor (CTRL + X)
-PLEX_URL="http://localhost:32400"
-# Follow guide here to retrieve the token
-# https://support.plex.tv/articles/204059436-finding-an-authentication-token-x-plex-token/
-PLEX_TOKEN=ChangeThisObviouslyFakePlexToken
+PLEX_URL="http://$plexaddress"
+PLEX_TOKEN=$plextoken
 PORT=8420
 ENV
-
-nano $mmatchDir/.env
 
 echo_progress_start "Installing systemd service and starting moviematch"
 cat > /etc/systemd/system/moviematch.service << SYSTEMD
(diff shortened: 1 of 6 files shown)


scripts/install/moviematch.sh
```diff
@@ -0,0 +1,70 @@
...
+[Unit]
+Description=Moviematch
+Documentation=https://swizzin.ltd/applications/moviematch
+After=network.target
+
+[Service]
+Type=simple
+User=moviematch
+WorkingDirectory=${mmatchDir}
+ExecStart=/usr/local/bin/deno run --allow-net --allow-read --allow-env ${mmatchDir}/src/index.ts
```
LukeChannings, line 48: The latest version also needs `--unstable` so that it can [handle SIGINT](https://deno.land/manual/examples/os_signals) correctly.

flying-sausages: Would this be sufficient?
```suggestion
ExecStart=/usr/local/bin/deno run --allow-net --allow-read --allow-env --unstable ${mmatchDir}/src/index.ts
```
I don't think it will necessarily cause issue on a seedbox if this application won't terminate "cleanly" so to speak. After all there is no database or any other writing to the disk. Is there something that this would add? I'll take your word for it :p

Thanks for having a look either way!

flying-sausages: Looks like it just doesn't run without it and restarts every 800ms lmao

flying-sausages:
Baseurl support added

scripts/update/moviematcher.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+#### Uncomment this when the base-url issue is resolved, and cahnge the nginx script
+
+# if [[ -f /install/.moviematch.lock ]]; then
+# 	if [[ -f /install/.nginx.lock ]]; then
+# 		bash /etc/swizzin/scripts/nginx/moviematch.sh
+# 	fi
+# fi
```
flying-sausages, line 9: Can be removed

scripts/install/moviematch.sh
```diff
@@ -0,0 +1,86 @@
...
+    . /etc/swizzin/sources/functions/deno
+    install_deno
+}
+
+mmatchDir="/opt/moviematch"
+
+_install_moviematch() {
+
+    echo_progress_start "Cloning moviematch"
+    git clone https://github.com/LukeChannings/moviematch.git $mmatchDir >> $log
```
flying-sausages, line 28: Should be switched to a release tarball download and an upgrader should be made for that

scripts/install/moviematch.sh
```diff
@@ -0,0 +1,86 @@
...
+    cat > $mmatchDir/.env << ENV
+PLEX_URL="http://$plexaddress"
+PLEX_TOKEN=$plextoken
+PORT=8420
+ENV
+}
+
+_systemd() {
+
+    echo_progress_start "Installing systemd service and starting moviematch"
```
flying-sausages, line 46: unfinished echo start

flying-sausages:
@LukeChannings Any chance i could ask you for a re-review? I am unable to get this running on the subpath, getting the output below all the time. Thanks for the help!

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <title translate="no">MovieMatch</title>
    <meta
      name="viewport"
      content="initial-scale=1, width=device-width, viewport-fit=cover"
    />
    <meta name="description" content="MovieMatch helps you pick a movie to watch on Plex" />
    <meta name="theme-color" content="#e64747" />
    <meta
      name="apple-mobile-web-app-status-bar-style"
      content="black-translucent"
    />
    <meta name="msapplication-TileColor" content="#e64747" />
    <link rel="icon" href="./icons/favicon.ico" />
    <link rel="icon" href="./icons/icon.svg" type="image/svg+xml" />
    <link rel="apple-touch-icon" href="./icons/icon-180.png" />
    <link rel="manifest" href="./manifest.webmanifest" />
    <link rel="stylesheet" href="./dist/main.css" />
    <script src="./dist/main.js" type="module" defer></script>
  </head>
  <body data-version="2.0.0-alpha.7
" data-base-path="/moviematch">
    <div id="app"></div>
  </body>
</html>
```

scripts/nginx/moviematch.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+cat > /etc/nginx/apps/moviematch.conf << NGINX
+    location ^~ /moviematch/ {
+        proxy_pass http://127.0.0.1:8420/;
+        proxy_set_header Upgrade \$http_upgrade;
```
LukeChannings, line 6: You can set the root path in the nginx config like:

`proxy_set_header x-forwarded-prefix /moviematch;`

scripts/nginx/moviematch.sh
```diff
@@ -0,0 +1,13 @@
...
+    }
+NGINX
+
+echo "host: 0.0.0.0
+basePath: /moviematch" >> /opt/moviematch/config.yaml
```
LukeChannings, line 11: `basePath` is renamed back to `rootPath` in the next release since I was getting too many bugs with people confused about the inconsistency between v1 and v2. Something to be aware of!

sources/functions/deno
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# Deno installer
+
+install_deno() {
```
LukeChannings, line 4: You don't need to install Deno if you're using the binary release, since it bundles a lightweight version of the Deno runtime.

LukeChannings (changes requested):
The markup you have looks correct to me (`data-root-path` is set correctly on the body).
I have some minor requests for how you're performing the root path config, but that shouldn't change the behaviour.

Do you get an error when looking at it in a browser?

PR #531, opened 2020-12-16
Update sabnzbd.sh

userdocs:
a working example of using a skeleton ini for configuring sab instead of using a start/shutdown/configure which is unreliable.

## Fixes issues: 
- na

## Proposed Changes:
- make sab configuration more consistent.
- 
## Categories

## Checklist

## Testing done
OS & Version: debian

### How have you tested this

Scenarios tested:
- I ran this and it worked
- I ran that and it didn't work

### DIFF
--- a/scripts/install/sabnzbd.sh
+++ b/scripts/install/sabnzbd.sh
@@ -68,30 +68,25 @@ Restart=on-failure
 
 [Install]
 WantedBy=multi-user.target
-
 SABSD
 
-systemctl enable -q --now sabnzbd 2>&1 | tee -a $log
-sleep 2
-echo_progress_done "SABnzbd started"
-
 echo_progress_start "Configuring SABnzbd"
-systemctl stop sabnzbd >> ${log} 2>&1
-sed -i "s/host_whitelist = .*/host_whitelist = $(hostname -f), $(hostname)/g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-sed -i "s|^host = .*|host = 0.0.0.0|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-sed -i "0,/^port = /s/^port = .*/port = 65080/" /home/${user}/.config/sabnzbd/sabnzbd.ini
-sed -i "s|^download_dir = .*|download_dir = ~/Downloads/incomplete|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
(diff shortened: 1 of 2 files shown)


scripts/install/sabnzbd.sh
```diff
@@ -9,52 +9,79 @@
...
-
-apt_install $LIST
+apt_install par2 p7zip-full python3-dev python3-setuptools python3-pip python3-venv libffi-dev libssl-dev libglib2.0-dev libdbus-1-dev
 
-python3_venv ${user} sabnzbd
+python3_venv "${user}" sabnzbd
 
-install_rar
+[[ "$(_os_distro)" == 'ubuntu' ]] && install_rar
+[[ "$(_os_distro)" == 'debian' ]] && _rar
```
flying-sausages, line 29: Why not handle the different cases for how to install rar in one place?

scripts/install/sabnzbd.sh
```diff
@@ -9,52 +9,79 @@
...
 
 echo_progress_start "Downloading and extracting sabnzbd"
-cd /opt
+cd /opt || exit 1
```
flying-sausages, line 32: I'd rather just get rid of the `cd` entirely and make the rest of these commands not be relative to current working dir

scripts/install/sabnzbd.sh
```diff
@@ -9,52 +9,79 @@
...
 mkdir -p /opt/sabnzbd
-wget -O sabnzbd.tar.gz $latest >> $log 2>&1
-tar xzf sabnzbd.tar.gz --strip-components=1 -C /opt/sabnzbd >> ${log} 2>&1
+_cmd_log wget -O sabnzbd.tar.gz "${latest}"
+_cmd_log tar xzf sabnzbd.tar.gz --strip-components=1 -C /opt/sabnzbd
```
flying-sausages, line 35: Re: `_cmd_log`, I'd take that out of this PR so we can just merge at as this will need a discussion on how to do this right and the rest of this should go in first

flying-sausages (changes requested):

scripts/install/sabnzbd.sh
flying-sausages: This should be reverted and should use #555 instead

flying-sausages:
FYI this is being reused in #590

PR #534, opened 2020-12-17
Fix venv paths that were mistakenly not updated during virtualenv migrations

liaralabs:
## Fixes issues: 
- `-bash: /opt/.venv/swizzin/bin/pip: /opt/swizzin/venv/bin/python3: bad interpreter: No such file or directory`
- after moving virtual environments, pip will not function because the env is still pointing to the wrong location

## Proposed Changes:
- `sed` the correct path into venvs if a fix is needed

## Categories

## Checklist

    - PR link: 
    - PR link: 

## Testing done
OS & Version:
grep and sed commands tested on Debian 10

### How have you tested this

Scenarios tested:
- Beyond sed and grep nothing at the moment

### DIFF
--- a/scripts/update/00-venv.sh
+++ b/scripts/update/00-venv.sh
@@ -16,10 +16,11 @@ if [[ -d /home/${user}/.venv ]]; then
 fi
 
 if [[ -d /opt/.venv ]]; then
-    envs=($(ls /opt/.venv))
-    for app in ${envs[@]}; do
-        if ! grep -q "#\!/opt/.venv/${app}/bin/python" /opt/.venv/${app}/bin/*; then
-            sed -i "s|#\!/.*/bin/python|#\!/opt/.venv/${app}/bin/python|g" /opt/.venv/${app}/bin/*
+    envs=($(find /opt/.venv/* -maxdepth 0 -type d))
+    for venvpath in ${envs[@]}; do
+        if ! grep -q "#\!${venvpath}/bin/python" ${venvpath}/bin/*; then
+            echo_log_only "Replacing venv path in $venvpath"
+            sed -i "s|#\!/.*/bin/python|#\!${venvpath}/bin/python|g" ${venvpath}/bin/*
         fi
     done
 fi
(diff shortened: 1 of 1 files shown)


scripts/update/00-venv.sh
```diff
@@ -14,3 +14,12 @@ if [[ -d /home/${user}/.venv ]]; then
         systemctl try-restart $app
     done
 fi
+
+if [[ -d /opt/.venv ]]; then
+    envs=($(ls /opt/.venv))
```
flying-sausages, line 19: We should start using find instead of ls as a rule of thumb

flying-sausages:
What are the steps to reproduce the broken scenario?

scripts/update/00-venv.sh
```diff
@@ -14,3 +14,12 @@ if [[ -d /home/${user}/.venv ]]; then
...
+    for app in ${envs[@]}; do
+        if ! grep -q "#\!/opt/.venv/${app}/bin/python" /opt/.venv/${app}/bin/*; then
+            sed -i "s|#\!/.*/bin/python|#\!/opt/.venv/${app}/bin/python|g" /opt/.venv/${app}/bin/*
+        fi
+    done
```
flying-sausages, line 24: I have not tested this at all so just shooting things on the wall
```suggestion
if [[ -d /opt/.venv ]]; then
    envs=($(find /opt/.venv/* -maxdepth 0 -type d))
    for venvpath in ${envs[@]}; do
        if ! grep -q "#\!/opt/.venv/${app}/bin/python" ${venvpath}/bin/*; then
            sed -i "s|#\!/.*/bin/python|#\!${venvpath}/bin/python|g" ${venvpath}/bin/*
        fi
    done
```

flying-sausages (changes requested):
This should future-proof in case a file ever ends up in `/opt/.venv` or something, and makes the seds a little easier to read

liaralabs:
> What are the steps to reproduce the broken scenario?

The panel is in this state right now for folks who are complaining that the update broke the network speeds/iowait indicators.

```
python3 -m venv foo
mv foo bar
./bar/bin/pip
```
Will create the scenario.

It can be fixed with a reinstall. I'm guessing this issue didn't come up with the prior move of venvs to opt because most applications never call pip again after moving (though other scripts in here have the erroneous shebang) and the python binary itself is not affected because it's a symlink.

PR #539, opened 2020-12-19
Lockfile updater

flying-sausages:
## Fixes issues: 
- It's annoying to maintain symlinks and whatnot just so your git doesn't fuck you with resets

## Proposed Changes:
- Add a flag to skip git pulls on updates called `--no-git`
- Add another method to skip the updates by doing `> /etc/swizzin/.dev.lock`

## Categories

- Changes structure of project (major rewrites, large directory changes, etc.)
- Impacts the flow of how we contribute (IDE changes, `.github` dir, etc.)

## Checklist

    - PR link: 
    - PR link: 

## Testing done
OS & Version:

### How have you tested this

image

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.dev.lock
--- a/scripts/box
+++ b/scripts/box
@@ -126,7 +126,7 @@ function _update() {
         echo_warn "Not updating /etc/swizzin as it is a symlink. Please consult your provider/maintainers in case you believe this is an error."
     elif [[ -e /etc/swizzin/.dev.lock ]]; then
         echo_warn "Not updating /etc/swizzin as a dev lockfile is present."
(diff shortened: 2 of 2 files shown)


liaralabs:
It seems to me the flag argument may conflict with the upgrade command, which should probably be a consideration.

flying-sausages:
> It seems to me the flag argument may conflict with the upgrade command, which should probably be a consideration.

The flag will indeed only work if it's right after upgrade, otherwise it will just try to run an upgrader for  --no-git. so consider it positional :p

i.e. `box upgrade --no-git lounge` will work but `box upgrade lounge --no-git` will not work as "expected" i suppose, but I don't think it's something we should spend time improving as this feature will be heavily **un-**documented

liaralabs (approved):

PR #542, opened 2020-12-20
feat: Readarr (Alpha) Installer

bakerboy448:
## Fixes issues: 
https://feathub.com/liaralabs/swizzin/+59

## Proposed Changes:
Create readarr installer
~~NOTE: Requires PR #544 merged prior to or with this PR~~
## Categories
- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

## Checklist -- TODO
- [X] Docs have been made OR are not necessary
    - PR link: https://github.com/liaralabs/docs.swizzin.ltd/pull/43
- [X] Changes to panel have been made OR are not necessary
    - PR link: https://github.com/liaralabs/swizzin_dashboard/pull/24
- [x] Code is formatted [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)
- [x] Shellcheck isn't screaming [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)
- [X] Prints to terminal are handled [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#printing-into-the-terminal)
- [X] I have commented my code, particularly in hard-to-understand areas

TODO: Testing
TODO: Migrate existing Radarr installs with conflicting and out of sync w/ upstream ssl port 8787 -> 9898
## Testing done
OS & Version:

### How have you tested this

Scenarios tested:
- I ran this and it worked
- I ran that and it didn't work

PR #544, opened 2020-12-20
misc(radarr): Various changes

### LINKED ISSUE #543: Radarr ports should be changed to default from upstream
non-issue upstream 9898 is default in v3
https://github.com/Radarr/Radarr/blob/develop/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs#L140

we should migrate any radarrs on 7878 to 9898...uh how? lol

_Originally posted by @bakerboy448 in https://github.com/swizzin/swizzin/pull/542#discussion_r546414320_

bakerboy448:
## Fixes issues: 
Fixes #543 

Prerequisite for #542 

## Proposed Changes:
Change Radarr SSL ports to match upstream

## Categories

## Checklist

TODO: Migrate Existing Installations

    - PR link: N/A
    - PR link: N/A

## Testing done
OS & Version: ubuntu 20.04

### How have you tested this

Scenarios tested:
- new ubuntu 
- installed radarr via swizzin
- ensure ssl port = 8787 and ssl - true
- box update works as expected


flying-sausages:
Btw is it just .net v3 that defaults to this port or is it both .net and mono?

We really need to figure out the appropriate time to change these ports, seeing as this will break a lot of installs for people. We might as well see if we can try API those changes back to some of these apps, but we absolutely need to be very vocal towards end-users about these ports changing when the updater runs.

bakerboy448:
> Btw is it just .net v3 that defaults to this port or is it both .net and mono?

Per the code, it is all. https://github.com/Radarr/Radarr/blob/develop/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs#L140

bakerboy448:
> see if we can try API those changes back to some of these apps

https://radarr.video/docs/api/#/Settings/put-config-host

should be easy enough!

liaralabs:
The SSL port shouldn't be in use either way. Isn't SSL port only used when SSL is true?

flying-sausages:
> shouldn't be in use

I'm considering the edge-case where someone is not using nginx but has a certificate and they added it to radarr themselves.

liaralabs:
Then we should just silently upgrade the port and only display a warning if SSL is true.  This is going to be a super edge case since you'd have to have technical skill to have changed the the config and generated the SSL certificate in the first place as this is not something our documentation discusses.

flying-sausages:
This needs an updater to do something like this
```
if sslport=8787; then
   sed sslport=9898 -i file
   if ssl=true;
      echo "port changed for ssl"
   fi
fi
```
lmk if you don't intend to do this, otherwise I'll jump on it

bakerboy448:
@flying-sausages go for it boss

bakerboy448:
sausage were you planning on knocking out the code to handle the extreme edge case?

(side note it seems pretty well SSL has material issues in the arrs and it doesn't seem to work based on bugs that have been seen....I'm not even sure if that edge case you wish to code for even exists in the wild)

bakerboy448:
Should be all set I think

scripts/update/radarr.sh
```diff
@@ -44,8 +44,17 @@ if [[ -f /install/.radarr.lock ]]; then
...
         elif [[ $(_radarr_version) = "mono-v2" ]]; then
-            echo_warn "Please upgrade your radarr to v3. It will be migrated to .Net core on the next \`box update\` run"
+            echo_warn "Radarr v0.2 is EOL and not supported. Please upgrade your radarr to v3. It will be migrated to .Net core on the next \`box update\` run"
             echo_docs "applications/radarr#migrating-to-v3-on-net-core"
         fi
     fi
+    #Mandatory SSL Port change for Readarr
+    sslport=$(grep -oPm1 "(?<=<SslPort>)[^<]+" /home/"$radarrOwner"/.config/Radarr/config.xml)
+    sslenabled=$(grep -oPm1 "(?<=<EnableSsl>)[^<]+" /home/"$radarrOwner"/.config/Radarr/config.xml)
+    if $sslport '==' '8787'; then
```
flying-sausages, line 54: not sure what that syntax was or if that would even work...
```suggestion
    if [ $sslport = '8787' ]; then
```

scripts/update/radarr.sh
```diff
@@ -44,8 +44,17 @@ if [[ -f /install/.radarr.lock ]]; then
...
+        sed s/\<SslPort\>port\=9898/\<SslPort\>port\=8787/ -i fil/home/"$radarrOwner"/.config/Radarr/config.xmle
+        if [ $sslenabled = true ]; then
+            echo "port changed for ssl due to possible future conflict with Readarr"
```
flying-sausages, line 57:
```suggestion
            echo_info "Radarr SSL port changed from 9898 to 8787, please ensure to adjust your dependent systems in case they were using this port."
```

scripts/update/radarr.sh
```diff
@@ -44,8 +44,17 @@ if [[ -f /install/.radarr.lock ]]; then
             fi
 
         elif [[ $(_radarr_version) = "mono-v2" ]]; then
-            echo_warn "Please upgrade your radarr to v3. It will be migrated to .Net core on the next \`box update\` run"
+            echo_warn "Radarr v0.2 is EOL and not supported. Please upgrade your radarr to v3. It will be migrated to .Net core on the next \`box update\` run"
```
flying-sausages, line 47: Might as well make this clear while we're here
```suggestion
            echo_warn "Radarr v0.2 is EOL and not supported. Please upgrade your radarr to v3. An attempt will be made to migrate to .Net core on the next \`box update\` run"
```

scripts/update/radarr.sh
```diff
@@ -11,19 +11,19 @@ if [[ -f /install/.radarr.lock ]]; then
...
-            echo_progress_done "Source downloaded"
+            echo_progress_done "Release downloaded"
 
             isactive=$(systemctl is-active radarr)
             echo_log_only "Radarr was $isactive"
             [[ $isactive == "active" ]] && systemctl stop radarr -q
 
-            echo_progress_start "Extracting archive"
-            rm /opt/Radarr/Radarr.exe
+            echo_progress_start "Extracting archive and removing old binaries"
```
flying-sausages, line 25: If we're nitpicking these echoes, "removing old binaries" is not at all what is being done here.

flying-sausages (changes requested):
See individual remarks. I will test this once the functional ones are resolved.

scripts/update/radarr.sh
```diff
@@ -44,8 +44,17 @@ if [[ -f /install/.radarr.lock ]]; then
...
-            echo_warn "Please upgrade your radarr to v3. It will be migrated to .Net core on the next \`box update\` run"
+            echo_warn "Radarr v0.2 is EOL and not supported. Please upgrade your radarr to v3. It will be migrated to .Net core on the next \`box update\` run"
             echo_docs "applications/radarr#migrating-to-v3-on-net-core"
         fi
     fi
+    #Mandatory SSL Port change for Readarr
+    sslport=$(grep -oPm1 "(?<=<SslPort>)[^<]+" /home/"$radarrOwner"/.config/Radarr/config.xml)
+    sslenabled=$(grep -oPm1 "(?<=<EnableSsl>)[^<]+" /home/"$radarrOwner"/.config/Radarr/config.xml)
+    if $sslport '==' '8787'; then
+        sed s/\<SslPort\>port\=9898/\<SslPort\>port\=8787/ -i fil/home/"$radarrOwner"/.config/Radarr/config.xmle
```
flying-sausages, line 55: what's `fil`?
```suggestion
        sed s/\<SslPort\>port\=9898/\<SslPort\>port\=8787/ -i /home/"$radarrOwner"/.config/Radarr/config.xmle
```

flying-sausages:
Actually, I will test this after you have tested this yourself, as there is no way this would have ran and then I'd spend time just picking off some low hanging fruit.

I appreciate starting PRs but if they're just being made to get us moving without being even checked whether the syntax works, I will just stop reviewing them all until there's some proof that they actually function.

scripts/update/radarr.sh
bakerboy448: This is exactly what is being done here; no action needed

we delete the old binaries in /opt/Radarr  and then extract the new tarball.

In fact these are from the official .net upgrade instructions
>Delete your existing binaries (contents or folder of /opt/Radarr) and replace with the contents of the .tar.gz you just downloaded.
>sudo mv /opt/Radarr /opt/Radarr.old Backup the old Binaries
>sudo rm -rf /opt/Radarr.old Remove the old binaries
https://wiki.servarr.com/Radarr_System#update_to_net_core_version

flying-sausages: `/opt/Radarr` is a directory, not a binary ;)

Plus, if we're doing this recursively, it would make sense to split them up in case the removal or the extrction take longer amounts of time to complete.

scripts/update/radarr.sh
flying-sausages: oh, it's `file` with the path pasted into it.

scripts/update/radarr.sh
bakerboy448: the directory contains the radarr binaries and hence that is what the development team refers to this as "removing the radarr binaries"

semantics aside, so what are you looking for here?  the whole folder must be removed and replaced.

bakerboy448:
tested; seems to work.

bakerboy448:
Fixes #639

scripts/install/radarr.sh
```diff
@@ -49,7 +49,7 @@ Group=${radarrOwner}
 Type=simple
 
 # Change the path to Radarr here if it is in a different location for you.
-ExecStart=/opt/Radarr/Radarr -nobrowser
+ExecStart=/opt/Radarr/Radarr -nobrowser -data=/home/${user}/.config/Radarr/
```
flying-sausages, line 52: is `$user` even used here? should be the owner var

scripts/update/radarr.sh
```diff
@@ -44,8 +57,20 @@ if [[ -f /install/.radarr.lock ]]; then
...
         fi
     fi
+    #Mandatory SSL Port change for Readarr
+    #shellcheck source=sources/functions/utils
+    . /etc/swizzin/sources/functions/utils
+    [[ -z $radarrOwner ]] && radarrOwner=$(_get_master_username)
+    sslport=$(grep -oPm1 "(?<=<SslPort>)[^<]+" /home/"$radarrOwner"/.config/Radarr/config.xml)
+    sslenabled=$(grep -oPm1 "(?<=<EnableSsl>)[^<]+" /home/"$radarrOwner"/.config/Radarr/config.xml)
+    if [[ "$sslport" = "8787" ]]; then
+        sed -i 's|<SslPort>8787</SslPort>|<SslPort>9898</SslPort>|g' "/home/$radarrOwner/.config/Radarr/config.xml"
```
flying-sausages, line 71:
```suggestion
        echo_log_only "Changing radarr ssl port"
        sed -i 's|<SslPort>8787</SslPort>|<SslPort>9898</SslPort>|g' "/home/$radarrOwner/.config/Radarr/config.xml"
```

flying-sausages (approved):

PR #438, opened 2020-10-27
Setup trimdown, Box command improvements

flying-sausages:
## Fixes issues: 
- Issue #380 
- Issues with weird names being used 

## Proposed Changes:
- Pulls down functions from github to let us use them from `setup.sh` from the very start
- De-duplicate a lot of the code between `box` and `setup.sh`
- Thoroughly handle all possible cases with the password and user checking
- Rewrite a lot of the user creation code

## Categories

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- Some existing `utils` got changed

## Checklist
- [x] Docs have been made/are not necessary
    - PR link: 
- [x] Changes to panel have been made/are not necessary
    - PR link: 
- [x] Code is formatted [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)
- [x] Shellcheck isn't screaming [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)
- [x] Prints to terminal are handled [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#printing-into-the-terminal)
- [x] I have commented my code, particularly in hard-to-understand areas

## Testing done
OS & Version: Ubuntu 20.04

### How have you tested this

Scenarios tested:
- [x] setup.sh flow
- [x] master user change password flow
- [x] new non-master user creation flow
- [x] non-master password change flow

PR #423, opened 2020-10-18
Making setup.sh unattended

flying-sausages:
skips all user interaction where possible, script updates to be compatible come later.

Documentation for this is available here https://github.com/liaralabs/docs.swizzin.ltd/pull/29

The docs for this are available here https://deploy-preview-29--swizzin-docs.netlify.app/guides/advanced-setup

PR #548, opened 2020-12-22
Simplified, unattended setup

flying-sausages:
This PR combines #438 and #423 with merge conflicts resolved so that we don't have to wait to merge them later


setup.sh
```diff
@@ -14,215 +14,228 @@
...
+            echo "/etc/swizzin/.dev.lock created"
+        fi
+        echo "Best of luck and please follow the contribution guidelines cheerio"
+    else
+        echo "Cloning swizzin repo to localhost"
+        git clone https://github.com/swizzin/swizzin.git /etc/swizzin >> ${log} 2>&1
+        echo -e "\tSwizzin cloned!"
+    fi
+    ln -s /etc/swizzin/scripts/ /usr/local/bin/swizzin
+    chmod -R 700 /etc/swizzin/scripts
```
liaralabs, line 49: Didn't you want to get rid of this line because it causes git to be chatty when doing `box update`s?

setup.sh
```diff
@@ -14,215 +14,228 @@
...
+        echo_progress_done
     fi
 
-    echo "Performing a system upgrade"
-    apt-get -q -y update >> ${log} 2>&1
-    apt-get -q -y upgrade >> ${log} 2>&1
+    apt_upgrade
+
+    if ! bash /etc/swizzin/scripts/update/10-dependencies.sh; then
+        echo_error "Dependencies failed to install"
```
liaralabs, line 191: Should we consider exiting here if depends fail? In my experience the install is going to botch pretty hard if these don't get installed.

scripts/box
```diff
@@ -304,19 +292,20 @@ function _adduser() {
...
     USER=${user^^}
-    cat > /etc/sudoers.d/${user} << SUD
+
+    #shellcheck source=sources/functions/utils
+    . /etc/swizzin/sources/functions/utils
+    cat > /etc/sudoers.d/"${user}" << SUD
 Cmnd_Alias   ${USER}CMDS = /bin/systemctl stop flood@${user}, /bin/systemctl restart flood@${user}, /bin/systemctl start flood@${user}, /bin/systemctl start rtorrent@${user}, /bin/systemctl stop rtorrent@${user}, /bin/systemctl restart rtorrent@${user}, /bin/systemctl start deluged@${user}, /bin/systemctl stop deluged@${user}, /bin/systemctl restart deluged@${user}, /bin/systemctl start deluge-web@${user}, /bin/systemctl stop deluge-web@${user}, /bin/systemctl restart deluge-web@${user}, /bin/systemctl start irssi@${user}, /bin/systemctl stop irssi@${user}, /bin/systemctl restart irssi@${user}, /bin/systemctl start qbittorrent@${user}, /bin/systemctl stop qbittorrent@${user}, /bin/systemctl restart qbittorrent@${user}, /usr/local/bin/swizzin/box chpasswd ${user}
 
 ${user}     ALL = (ALL) NOPASSWD: ${USER}CMDS
 SUD
```
liaralabs, line 302: We don't want this to run if the user is master as they get full sudo permissions

liaralabs (changes requested):
So it looks like we still need unattended variables for:
- qBittorrent
- Libtorrent
- Deluge
- rTorrent

Unless these are not high priority since the installer will ask questions off the bat; however, for it to be truly unattended we should have a mechanism in place for the user to be able to skip these prompts. I don't know if we want to wait until the libtorrent overhaul takes place, but I'd like to get this PR merged sooner rather than later.

liaralabs:
I also think we could push this to the next level by having the ability to define a post-install script

setup.sh
flying-sausages: Not if we keep our permissions in check ha

scripts/box
flying-sausages: Aren't these the sudo permissions to run their own tasks?

PR #555, opened 2020-12-28
Unify rar installation across distros and arches

flying-sausages:
## Fixes issues: 
- Ubuntu has `rar` in sources, Debian has to have a repo enabled, and `arm64` is a whole other mess
- Installs are killed if `rar` is not available

## Proposed Changes:
- Unify function for `rar` installation
- Explicitly handle each separate OS-Arch case
- Build `unrar` from source on un-handled architectures
- Explicitly mention when `rar` might not be available **but allow installations to continue**
- ~~Manage `rar` install status through lockfile~~
- ~~Add installer for `rar` (and `unrar`)~~
- ~~Add remover for `rar` (and `unrar`)~~

## Categories

- ~~New feature (non-breaking change which adds functionality)~~
- Some existing `functions` got changed (e.g. `utils`, `os`, `apt`, `ask`, etc.)

## Architectures

- `x86_64` is specifically handled as:
    - [x] Success
    - [ ] Error
    - Notes: Largely unchanged
- `arm64` is specifically handled as:
    - [x] Success
    - [ ] Error
    - Notes: `unrar` only (either via `apt` or installed from source)
- [x] Changes are arch agnostic 
    - Notes: `unrar` installed from source if no other combination is found

## Checklist

    - PR link: 
    - PR link: 

## Testing done
OS & Version: Ubuntu for arm

### How have you tested this

`unrar` installs fine on `arm64` ubuntu
`unrar` builds from source on the same arch too


liaralabs:
So I'm looking this over because in the current state, sabnzbd will fail to install rar.

That said, I don't necessarily agree with this being its own separate package entirely given that this package is dependency, but perhaps other folks have opinions on this. However, the package list is growing *insanely* long and this just further clutters it up with an application which should just be seamlessly installed if required by a package.

flying-sausages:
Alright, I'll remove the installers and leave it as a function.

flying-sausages:
> However, the package list is growing _insanely_ long and this just further clutters it up with an application which should just be seamlessly installed if required by a package.

Actually the way this is designed, the same function will just quietly install if the dependency is missing. The same function that installers called before is now handling more cases. Nothing will fail if it requires `rar` and the `rar` package isn't installed beforehand.

It just opens the paths to remove the package, and install it without having to install the depndee for it. This would allow OSs without `rar`/`unrar` packaged to install it.

flying-sausages:
I take the last comments back lol let's just not change the way it's managed rn

flying-sausages:
Success on ubuntu for arm when installing rutorrent, should be tested on debian x64 still and then it's good to merge
afbeelding

scripts/update/10-dependencies.sh
```diff
@@ -17,3 +17,19 @@ if [[ ${missing[0]} != "" ]]; then
...
+    fi
+    if ! grep -q non-free /etc/apt/sources.list; then
+        echo_info "Enabling non-free repo"
+        apt-add-repository non-free >> ${log} 2>&1
+        trigger_apt_update=true
+    fi
+    if [[ $trigger_apt_update == "true" ]]; then
+        apt_update
+    fi
+fi
```
flying-sausages, line 35: Should this be above the actual dependency installation?

liaralabs: `apt-add-repository` will not be present on most Debian installs, thus it needs to be done after the deps are checked. I don't think any of the global deps themselves should be coming from non-free or contrib

flying-sausages: 👍 thanks

PR #590, opened 2021-02-09
Update sabnzbd and futureproof

liaralabs:
## Description
- Implements a new python environment management tactic
- Gets sabnzbd ready for v3.2
- Uses userdocs implementation of a template to minimize problems when setting up sabnzbd for the first time
- New update flow

<details>
<summary>Label of collapsible goes here</summary>

</details>

## Fixes issues: 
```
sed: can't read /home/username/.config/sabnzbd/sabnzbd.ini: No such file or directory
```

## Change Categories

- Bug fix 

## Checklist

    - PR link: 
    - PR link: 

### Architectures

| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|

### DIFF
--- a/scripts/install/sabnzbd.sh
+++ b/scripts/install/sabnzbd.sh
@@ -49,12 +49,12 @@ install_rar
 
 echo_progress_start "Downloading and extracting sabnzbd"
 mkdir -p /opt/sabnzbd
-wget -O /opt/sabnzbd.tar.gz "$latest" >> "${log}" 2>&1 || {
+wget -O /tmp/sabnzbd.tar.gz "$latest" >> "${log}" 2>&1 || {
     echo_error "Failed to download archive"
     exit 1
 }
-tar xzf /opt/sabnzbd.tar.gz --strip-components=1 -C /opt/sabnzbd >> "${log}" 2>&1
-rm -rf /opt/sabnzbd.tar.gz
+tar xzf /tmp/sabnzbd.tar.gz --strip-components=1 -C /opt/sabnzbd >> "${log}" 2>&1
+rm -rf /tmp/sabnzbd.tar.gz
 echo_progress_done
 
 echo_progress_start "Installing pip requirements"
--- a/scripts/install/sabnzbd.sh
+++ b/scripts/install/sabnzbd.sh
@@ -32,6 +32,7 @@ else
 fi
 
 apt_install $LIST
+install_rar
 
 case ${PYENV} in
     True)
@@ -44,8 +45,6 @@ case ${PYENV} in
         python3_venv ${user} sabnzbd
         ;;
 esac
-
-install_rar
 
 echo_progress_start "Downloading and extracting sabnzbd"
 mkdir -p /opt/sabnzbd
--- a/scripts/install/sabnzbd.sh
+++ b/scripts/install/sabnzbd.sh
@@ -41,6 +41,8 @@ echo_progress_start "Installing pip requirements"
 if [[ $latestversion =~ ^3\.0\.[1-2] ]]; then
     sed -i "s/feedparser.*/feedparser<6.0.0/g" /opt/sabnzbd/requirements.txt
 fi
+
+/opt/.venv/sabnzbd/bin/pip install --upgrade pip wheel >> "${log}" 2>&1
 /opt/.venv/sabnzbd/bin/pip install -r /opt/sabnzbd/requirements.txt >> "${log}" 2>&1
 echo_progress_done
 
--- a/scripts/install/sabnzbd.sh
+++ b/scripts/install/sabnzbd.sh
@@ -84,7 +84,6 @@ Restart=on-failure
 
 [Install]
 WantedBy=multi-user.target
-
 SABSD
 
 systemctl enable -q --now sabnzbd 2>&1 | tee -a $log
@@ -104,19 +103,23 @@ echo_progress_done "SABnzbd started"
 
 echo_progress_start "Configuring SABnzbd"
 systemctl stop sabnzbd >> ${log} 2>&1
-sed -i "s/host_whitelist = .*/host_whitelist = $(hostname -f), $(hostname)/g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-#sed -i "s|^host = .*|host = 0.0.0.0|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-sed -i "0,/^port = /s/^port = .*/port = 65080/" /home/${user}/.config/sabnzbd/sabnzbd.ini
-sed -i "s|^download_dir = .*|download_dir = ~/Downloads/incomplete|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-sed -i "s|^complete_dir = .*|complete_dir = ~/Downloads/complete|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-#sed -i "s|^ionice = .*|ionice = -c2 -n5|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-#sed -i "s|^par_option = .*|par_option = -t4|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-#sed -i "s|^nice = .*|nice = -n10|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-#sed -i "s|^pause_on_post_processing = .*|pause_on_post_processing = 1|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-#sed -i "s|^enable_all_par = .*|enable_all_par = 1|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-#sed -i "s|^direct_unpack_threads = .*|direct_unpack_threads = 1|g" /home/${user}/.config/sabnzbd/sabnzbd.ini
-sed -i "0,/password = /s/password = .*/password = ${password}/" /home/${user}/.config/sabnzbd/sabnzbd.ini
-sed -i "0,/username = /s/username = .*/username = ${user}/" /home/${user}/.config/sabnzbd/sabnzbd.ini
+cat > /home/${user}/.config/sabnzbd/sabnzbd.ini << SAB_INI
+[misc]
+host_whitelist = $(hostname -f), $(hostname)
+host = 0.0.0.0
+port = 65080
+download_dir = ~/Downloads/incomplete
+complete_dir = ~/Downloads/complete
+ionice = -c2 -n5
+par_option = -t4
+nice = -n10
+pause_on_post_processing = 1
+enable_all_par = 1
+direct_unpack_threads = 1
+password = "${password}"
+username = "${user}"
+SAB_INI
+
 systemctl restart sabnzbd >> ${log} 2>&1
 echo_progress_done
 
--- a/scripts/install/sabnzbd.sh
+++ b/scripts/install/sabnzbd.sh
@@ -14,7 +14,7 @@
 . /etc/swizzin/sources/functions/utils
 
 user=$(_get_master_username)
-password=$(_get_user_password ${user})
+password="$(_get_user_password ${user})"
 #latest=$(curl -s https://sabnzbd.org/downloads | grep -m1 Linux | grep download-link-src | grep -oP "href=\"\K[^\"]+")
 latest=$(curl -sL https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest | jq -r '.assets[]?.browser_download_url | select(contains("tar.gz"))') || {
     echo_error "Failed to query GitHub for latest sabnzbd version"
--- a/scripts/install/sabnzbd.sh
+++ b/scripts/install/sabnzbd.sh
@@ -20,6 +20,7 @@ latest=$(curl -sL https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest |
 latestversion=$(echo $latest | awk -F "/" '{print $NF}' | cut -d- -f2)
 systempy3_ver=$(get_candidate_version python3)
 
+#Version 3.2 is going to raise the min python version to 3.6 so we have to differentiate whether or not to build a pyenv
 if dpkg --compare-versions ${systempy3_ver} lt 3.6.0 && dpkg --compare-versions ${latestversion} ge 3.2.0; then
     LIST='par2 p7zip-full libffi-dev libssl-dev libglib2.0-dev libdbus-1-dev'
     PYENV=True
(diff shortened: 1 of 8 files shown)


flying-sausages:
Which environments did this get tested on?

Also, looks like for some reason not all flags from `shfmt` are being applied so it has to be applied manually right now. I'm considering switching the set up to use Prettier.io which might be a bit more overkill but at least it has traffic and is supported...

scripts/install/sabnzbd.sh
```diff
@@ -10,30 +10,45 @@
...
+        pyenv_install_version 3.7.7
+        pyenv_create_venv 3.7.7 /opt/.venv/sabnzbd
+        chown -R ${user}: /opt/.venv/sabnzbd
+        ;;
+    *)
+        python3_venv ${user} sabnzbd
+        ;;
+esac
 
 install_rar
```
flying-sausages, line 45: Can we either get rid of this from the installer or use #555 for it?

scripts/install/sabnzbd.sh
```diff
@@ -10,30 +10,45 @@
...
-
 . /etc/swizzin/sources/functions/pyenv
 . /etc/swizzin/sources/functions/utils
 
-LIST='par2 p7zip-full python3-dev python3-setuptools python3-pip python3-venv libffi-dev libssl-dev libglib2.0-dev libdbus-1-dev'
+user=$(_get_master_username)
+password=$(_get_user_password ${user})
+#latest=$(curl -s https://sabnzbd.org/downloads | grep -m1 Linux | grep download-link-src | grep -oP "href=\"\K[^\"]+")
+latest=$(curl -sL https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest | grep -Po '(?<="browser_download_url":).*?[^\\].tar.gz"' | sed 's/"//g')
+latestversion=$(echo $latest | awk -F "/" '{print $NF}' | cut -d- -f2)
```
flying-sausages, line 20:
```suggestion
latest=$(https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest | jq -r .tarball_url)
latestversion=$(curl https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest | jq -r .tag_name)
```

scripts/install/sabnzbd.sh
```diff
@@ -10,30 +10,45 @@
...
-#latest=$(curl -s https://sabnzbd.org/downloads | grep -m1 Linux | grep download-link-src | grep -oP "href=\"\K[^\"]+")
-latest=$(curl -sL https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest | grep -Po '(?<="browser_download_url":).*?[^\\].tar.gz"' | sed 's/"//g')
-latestversion=$(echo $latest | awk -F "/" '{print $NF}' | cut -d- -f2)
-
 . /etc/swizzin/sources/functions/pyenv
 . /etc/swizzin/sources/functions/utils
 
-LIST='par2 p7zip-full python3-dev python3-setuptools python3-pip python3-venv libffi-dev libssl-dev libglib2.0-dev libdbus-1-dev'
+user=$(_get_master_username)
+password=$(_get_user_password ${user})
```
flying-sausages, line 17:
```suggestion
password="$(_get_user_password ${user})"
```

scripts/install/sabnzbd.sh
```diff
@@ -10,30 +10,45 @@
...
+esac
 
 install_rar
 
 echo_progress_start "Downloading and extracting sabnzbd"
 cd /opt
 mkdir -p /opt/sabnzbd
-wget -O sabnzbd.tar.gz $latest >> $log 2>&1
-tar xzf sabnzbd.tar.gz --strip-components=1 -C /opt/sabnzbd >> ${log} 2>&1
+wget -O sabnzbd.tar.gz $latest >> "${log}" 2>&1
```
flying-sausages, line 50:
```suggestion
wget -O sabnzbd.tar.gz "$latest" >> "${log}" 2>&1
```

scripts/update/sabnzbd.sh
```diff
@@ -7,7 +7,7 @@ if [[ -f /install/.sabnzbd.lock ]]; then
         codename=$(lsb_release -cs)
```
flying-sausages, line 7: the could use an `_os` function

scripts/upgrade/sabnzbd.sh
```diff
@@ -1,40 +1,63 @@
...
 if [[ ! -f /install/.sabnzbd.lock ]]; then
     echo_error "SABnzbd not detected. Exiting!"
     exit 1
 fi
 
+. /etc/swizzin/sources/functions/pyenv
 localversion=$(/opt/.venv/sabnzbd/bin/python /opt/sabnzbd/SABnzbd.py --version | grep -m1 SABnzbd | cut -d- -f2)
 #latest=$(curl -s https://sabnzbd.org/downloads | grep -m1 Linux | grep download-link-src | grep -oP "href=\"\K[^\"]+")
 latest=$(curl -sL https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest | grep -Po '(?<="browser_download_url":).*?[^\\].tar.gz"' | sed 's/"//g')
 latestversion=$(echo $latest | awk -F "/" '{print $NF}' | cut -d- -f2)
```
flying-sausages, line 13: Can re-use suggestion from above

scripts/upgrade/sabnzbd.sh
```diff
@@ -1,40 +1,63 @@
...
+        if grep -q python2 /etc/systemd/system/sabnzbd.service; then
+            sed -i 's/python2/python/g' /etc/systemd/system/sabnzbd.service
+            systemctl daemon-reload
+        fi
+
+        echo_progress_done
+    fi
     cd /opt/
     echo_progress_start "Downloading latest source"
     wget -q -O sabnzbd.tar.gz $latest
```
flying-sausages, line 49:
```suggestion
    wget -q -O sabnzbd.tar.gz "$latest"
```

scripts/upgrade/sabnzbd.sh
```diff
@@ -1,40 +1,63 @@
...
+            python3_venv ${user} sabnzbd
+        fi
+        if grep -q python2 /etc/systemd/system/sabnzbd.service; then
+            sed -i 's/python2/python/g' /etc/systemd/system/sabnzbd.service
+            systemctl daemon-reload
+        fi
+
+        echo_progress_done
+    fi
     cd /opt/
```
flying-sausages, line 47: Any chance we get rid of the `cd`s?

Kwang-hub:
Hello 

I was him who have this issue: 

sed: can't read /home/unnamed/.config/sabnzbd/sabnzbd.ini: No such file or directory

Will you contact me, when it is ready to install on " Swizzin " - on "Swizzin-box"

scripts/install/sabnzbd.sh
flying-sausages: This will of course only work if the tag tarball is the same as the relase file ending in .tar.gz, so if the released source is modified compared to the repo, we should target the actual release instead

scripts/install/sabnzbd.sh
```diff
@@ -10,30 +10,45 @@
...
+tar xzf sabnzbd.tar.gz --strip-components=1 -C /opt/sabnzbd >> "${log}" 2>&1
```
flying-sausages, line 51:
```suggestion
wget -O /tmp/sabnzbd.tar.gz $latest >> "${log}" 2>&1
tar xzf /tmp/sabnzbd.tar.gz --strip-components=1 -C /opt/sabnzbd >> "${log}" 2>&1
```

scripts/install/sabnzbd.sh
```diff
@@ -10,30 +10,45 @@
...
 rm -rf sabnzbd.tar.gz
```
flying-sausages, line 52: worth removing the `cd` and full-pathing the archive location

scripts/install/sabnzbd.sh
liaralabs: I'd prefer to avoid `curl`ing the api twice. It takes extra time to make another request and also uses another (limited) request. The `latest` tag should have all the information we need to properly set the `latestversion` variable

liaralabs:
> Will you contact me, when it is ready to install on " Swizzin " - on "Swizzin-box"

Subscribe to the issue and you will receive all updates directly from GitHub.

scripts/install/sabnzbd.sh
liaralabs: Also 
```
curl -sL https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest | jq -r .tarball_url
```
Does not return a `.tar.gz` link

liaralabs: The `tarball.url` is literally the entire repo tar'd at time of release and is not the same as the src tar gz

image

flying-sausages: > I'd prefer to avoid `curl`ing the api twice

Sure, we can store the first response, the point was mostly to switch to using `jq` instead which is cleaner and more expectable

> Does not return a `.tar.gz` link

afbeelding


> The `tarball.url` is literally the entire repo tar'd at time of release and is not the same as the src tar gz

Fair enough, I'll come back to you on that with a proper `jq` filter to retrieve the released .tar.gz

liaralabs: Already got it ;)

```
latest=$(curl -sL https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest | jq -r '.assets[]?.browser_download_url | select(contains("tar.gz"))')
latestversion=$(echo $latest | awk -F "/" '{print $NF}' | cut -d- -f2)
```

scripts/upgrade/sabnzbd.sh
liaralabs: What harm is the `cd` causing here? I'd rather the tarball get downloaded into opt than `cwd`

scripts/install/sabnzbd.sh
```diff
@@ -10,30 +10,51 @@
...
 
-LIST='par2 p7zip-full python3-dev python3-setuptools python3-pip python3-venv libffi-dev libssl-dev libglib2.0-dev libdbus-1-dev'
+user=$(_get_master_username)
+password=$(_get_user_password ${user})
+#latest=$(curl -s https://sabnzbd.org/downloads | grep -m1 Linux | grep download-link-src | grep -oP "href=\"\K[^\"]+")
+latest=$(curl -sL https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest | jq -r '.assets[]?.browser_download_url | select(contains("tar.gz"))') || {
+    echo_error "Failed to query GitHub for latest sabnzbd version"
+    exit 1
+}
+latestversion=$(echo $latest | awk -F "/" '{print $NF}' | cut -d- -f2)
```
flying-sausages, line 23:
```suggestion
ret=$(curl -sL https://api.github.com/repos/sabnzbd/sabnzbd/releases/latest) || {
    echo_error "Failed to query GitHub"
    exit 1
}
latest=$(echo $ret | jq -r '.assets[]?.browser_download_url | select(contains("tar.gz"))') 
latestversion=$(echo $ret | jq -r .tag_name)
```
`jq` > `awk` & `cut`ing `JSON`s

liaralabs: It's not `awk`'ing json if you closely inspect the output of `latest`. It's parsing the URL itself.

scripts/upgrade/sabnzbd.sh
flying-sausages: the `wget` can still do something like `-O /tmp/archive.tar.gz`. Downloading to `cwd` is a sin, and `cd` being successful should not be taken for granted

scripts/install/sabnzbd.sh
flying-sausages: Ah yes I didn't look closely enough

liaralabs: Furthermore, I believe my current implementation of grabbing the full url in one command is more resilient as your command will only fail if the API is down, but the committed command will fail if a `.tar.gz` link is not returned.

flying-sausages: Truth!

scripts/install/sabnzbd.sh
liaralabs: I don't want a pending PR to block the merging of this script.

scripts/upgrade/sabnzbd.sh
liaralabs: > cd being successful should not be taken for granted
True but we also create the directory in the directly preceding command. Shellcheck is not god and it's suggestions have led me astray from time to time 😉

flying-sausages: I'm suggesting this
```bash
wget $link -O /tmp/archive.tar.gz
tar xzf /tmp/archive.targz -C /opt/sbanznznzd --strip
rm /tmp/archive.tar.gz
```
I'm not even referring to shellcheck here, `cd`ing and manipulating `cwd` is just not a great idea imo

scripts/upgrade/sabnzbd.sh
```diff
@@ -47,13 +47,15 @@ if dpkg --compare-versions ${localversion} lt ${latestversion}; then
...
-    cd /opt/
     echo_progress_start "Downloading latest source"
-    wget -q -O sabnzbd.tar.gz $latest
-    rm -rf sabnzbd/*
-    sudo -u ${user} bash -c "tar xzf sabnzbd.tar.gz --strip-components=1 -C /opt/sabnzbd" >> "$log" 2>&1
+    wget -q -O /tmp/sabnzbd.tar.gz "$latest"
+    rm -rf /opt/sabnzbd/*
+    if [[ ! -d /opt/sabnzbd ]]; then
+        mkdir -p /opt/sabnzbd
+    fi
```
flying-sausages, line 55:
```suggestion
    rm -rf /opt/sabnzbd/
    mkdir -p /opt/sabnzbd
```
maybe a little simpler? not sure whether there's some edge-case this is trying to catch

liaralabs:
> 
> 
> Which environments did this get tested on?
> 
> Also, looks like for some reason not all flags from `shfmt` are being applied so it has to be applied manually right now. I'm considering switching the set up to use Prettier.io which might be a bit more overkill but at least it has traffic and is supported...

Not sure what's going on here as my shfmt arguments are exactly the same as the ones in `.vscode/settings.json`

flying-sausages:
> Not sure what's going on here as my shfmt arguments are exactly the same as the ones in `.vscode/settings.json`

I ran into this recently for some reason too. The formatting on-save was not applying the `-sr` flag properly so I had to manually `shfmt` the whole thing before pushing. Something something vscode plugin being broken I think

PR #595, opened 2021-02-23
Libtorrent overhaul for static builds

liaralabs:
## Description
Update libtorrent builds to use static binaries rather than shared objects. Allows things like Deluge 1.3.15 coexisting with qBit 4.3.3 as libtorrent is built in to the respective applications

## Fixes issues: 
- Libtorrent 1.2.11 not yet supported.
- pip2.7 has been band-aid fixed to allow focal to keep building Deluge 1.3

## Proposed Changes:
- Simplify libtorrent-based application code
- monolithic repo/compile option (can be changed later) affects all libtorrent installs
- Builds will now optimize based on the processor features (`-march=native`)
- b2 builds for libtorrent means libboost no longer needs to be installed which means focal doesn't need to maintain 2 boost lib versions
  - No longer allow the user to choose libtorrent versions. They are picked for the user depending on the client and version.
- Deprecate the ability to use shared libtorrent libraries if compile options are set

## Change Categories

- New feature 
- Change in `functions` 
- Project structure 

## Checklist

    - PR link: 
    - PR link: 

## Test scenarios

What's been tested:
Debian 10:
✅ Install dynamic libs from master and upgrade to static or repo on overhaul branch
✅ Install repo libs and upgrade to repo/compile
✅ Fresh install

Ubuntu 20.04:
✅ Fresh install

ARM HAS NOT BEEN TESTED (I cannot)

### Architectures

| Focal 		|	🛠		|			|	🛠		|				|
| Bionic		|			|			|			|				|
| Buster		|	✅		|			|			|				|
| Stretch		|			|			|			|				|

### 🛠🛠 TODO
- Unattended install support 
  - LIBTORRENT_RASTERBAR_METHOD already exists and should be exportable on setup (compile or repo)
  - Need deluge versioning
  - qBit versioning


sources/functions/qbittorrent
```diff
@@ -41,48 +36,110 @@ function whiptail_qbittorrent() {
...
-        else
-            LIST='build-essential pkg-config automake libtool git zlib1g-dev libssl-dev libgeoip-dev libboost-dev libboost-system-dev libboost-chrono-dev libboost-random-dev python3 qtbase5-dev qttools5-dev-tools zlib1g-dev'
-        fi
+        LIST='build-essential automake pkg-config libtool git zlib1g-dev libssl-dev libgeoip-dev python3 qtbase5-dev qttools5-dev-tools zlib1g-dev'
+
         apt_install --no-recommends $LIST
         cd /tmp
         rm_if_exists "/tmp/qbittorrent"
-        wget -q https://github.com/qbittorrent/qBittorrent/archive/release-${VERSION}.tar.gz >> "${log}" 2>&1
+        wget https://github.com/qbittorrent/qBittorrent/archive/release-${VERSION}.tar.gz >> "${log}" 2>&1
```
flying-sausages, line 123: Please let's make these full path and not pwd relative while this is being changed

flying-sausages (commented):
Very brief read-through, will try run through this some other time

sources/functions/qbittorrent
```diff
@@ -41,48 +36,110 @@ function whiptail_qbittorrent() {
...
+            cryptography_type=built-in
+            ;;
+        4.4.*)
+            libtorrent_branch=RC_2_0
+            libtorrent_minimum_version=2.0
+            stdcver="c++17"
+            cryptography_type=built-in
+            ;;
+        *)
+            echo_error "qBittorrent version should be between 4.1.*.* and 4.3.*.* -- Setup will not proceed."
```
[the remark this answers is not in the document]
liaralabs: 4.4 does not currently exist, this code is here only for preparation purposes. It would be foolish to suggest to people that it currently a valid option.

I don't see a way to fix your issue with "4.1.0.0" as versions like 4.1.9.1 exist though they are rare. Throwing regex into the error message will only serve to confuse people. I'm assuming that folks will be looking at a list of releases and choosing the version number they are looking for if they are not using menus.

sources/functions/qbittorrent
```diff
@@ -15,21 +14,17 @@ function whiptail_qbittorrent() {
...
-                    "Repo" "(${repov})" \
                     "4.1" "(${latestv41})" 3>&1 1>&2 2>&3)
                 ;;
             *)
                 function=$(whiptail --title "Install Software" --menu "Choose a qBittorrent version:" --ok-button "Continue" --nocancel 12 50 4 \
-                    "Repo" "(${repov})" \
                     "4.1" "(${latestv41})" \
                     "4.2" "(${latestv42})" \
                     "4.3" "(${latestv43})" 3>&1 1>&2 2>&3)
                 #"Latest" "(${latestv})" 3>&1 1>&2 2>&3)
```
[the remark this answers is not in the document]
liaralabs: No, the option is now baked in above. If you change the method between repo and compile you will need to `export LIBTORRENT_RASTERBAR_METHOD=repo` before running the upgrade script (this will need to be clarified in the docs).  I don't see people switching between repo and compile often once chosen so I don't want this popup running every time.

scripts/install/deluge.sh
```diff
@@ -13,267 +13,14 @@
...
-    fi
-
-    touch /install/.deluge.lock
-    touch /install/.delugeweb.lock
-}
-
 . /etc/swizzin/sources/functions/deluge
 . /etc/swizzin/sources/functions/libtorrent
 . /etc/swizzin/sources/functions/utils
+. /etc/swizzin/sources/functions/fpm
```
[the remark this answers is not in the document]
liaralabs: I don't see the harm in including it globally. It's easier to see what the script requires if everything is included in one section.

sources/functions/libtorrent
```diff
@@ -1,123 +1,47 @@
...
-        if [[ $function == Repo ]]; then
-            export libtorrent=repo
-        elif [[ $function == 1.0 ]]; then
-            export libtorrent=RC_1_0
-        elif [[ $function == 1.1 ]]; then
-            export libtorrent=RC_1_1
-        elif [[ $function == 1.2 ]]; then
-            export libtorrent=RC_1_2
+            echo_info "Install will now exit"
+            exit 1
```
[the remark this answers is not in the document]
liaralabs: Yep, the rest of the code is currently un-used. I left it in the event I decide to make changes in the future, however the message above states that the install will now exit, which it does.

sources/functions/libtorrent
```diff
@@ -126,77 +50,70 @@ function check_client_compatibility() {
...
+        qbittorrent=$(qbittorrent-nox --version 2> /dev/null | grep -oP '\d+\.\d+\.\d+')
+    fi
+    apt_remove "${conflicting_libtorrent_packages[@]}"
+    if [[ -n $deluge && ! ${called_by} == deluge && "${conflicting_libtorrent_packages[*]}" =~ python ]]; then
+        #shellcheck source=sources/functions/deluge
+        . /etc/swizzin/sources/functions/deluge
+        . /etc/swizzin/sources/functions/fpm
+        echo_info "Recompiling Deluge for compatibility before continuing"
+        install_fpm
+        apt_remove --purge ^deluge.*
```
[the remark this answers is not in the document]
liaralabs: Shouldn't need to be. I would encourage you to test this behavior for yourself if you don't believe it.

sources/functions/qbittorrent
```diff
@@ -41,48 +36,110 @@ function whiptail_qbittorrent() {
...
+    false
+    return
+}
+
+function build_libtorrent_qbittorrent() {
+    booststrap
+    echo "using gcc : : : <cflags>\"-march=native -std=${stdcver}\" <cxxflags>\"-march=native -std=${stdcver}\" ;" > /root/user-config.jam
+
+    cd /tmp
+    git clone -b ${libtorrent_branch} https://github.com/arvidn/libtorrent >> ${log} 2>&1
```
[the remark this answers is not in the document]
liaralabs: The branch tag? Yes the `-b` tag on this git clone has been in use since the project's beginning.

sources/functions/qbittorrent
```diff
@@ -41,48 +36,110 @@ function whiptail_qbittorrent() {
...
+    fi
+    false
+    return
+}
+
+function build_libtorrent_qbittorrent() {
+    booststrap
+    echo "using gcc : : : <cflags>\"-march=native -std=${stdcver}\" <cxxflags>\"-march=native -std=${stdcver}\" ;" > /root/user-config.jam
+
+    cd /tmp
```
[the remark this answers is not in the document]
liaralabs: Regretfully, I don't think `cd` can be avoided where `b2` is involved

sources/functions/libtorrent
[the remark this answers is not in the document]
flying-sausages: If that's the case I'd comment it out just in case the `exit` would get removed or something, it's slightly confusing imo

sources/functions/qbittorrent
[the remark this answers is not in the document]
liaralabs: I've removed some `cd`s but not all can be avoided

sources/functions/libtorrent
```diff
@@ -207,153 +124,60 @@ function whiptail_skip_libtorrent_rasterbar() {
...
+function check_libtorrent_rasterbar_method() {
+    if [[ -n $LIBTORRENT_RASTERBAR_METHOD ]] && [[ ! $(swizdb get "libtorrent_rasterbar_method") == "${LIBTORRENT_RASTERBAR_METHOD}" ]]; then
+        swizdb set "libtorrent_rasterbar_method" "${LIBTORRENT_RASTERBAR_METHOD}"
+    fi
+    if ! swizdb get "libtorrent_rasterbar_method" > /dev/null 2>&1; then
+        whiptail_libtorrent_rasterbar_method
+    fi
+    if [[ -z ${LIBTORRENT_RASTERBAR_METHOD} ]]; then
+        LIBTORRENT_RASTERBAR_METHOD=$(swizdb get libtorrent_rasterbar_method)
+    fi
```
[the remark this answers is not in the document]
liaralabs: If you can think of better logic, then sure. However, all these cases are quite separate

sources/functions/qbittorrent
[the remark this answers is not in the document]
flying-sausages: we can probably check for validity of the version before the switchcase with a function that could check the tags available upstream or in the git repository. Then the default case would be more like "Version is valid but not currently handled, please raise a ticket or proceed with the last version's variables"

sources/functions/qbittorrent
flying-sausages:
```suggestion
    git clone -b ${libtorrent_branch} https://github.com/arvidn/libtorrent /tmp/libtorrent >> ${log} 2>&1 || {
        echo_error "Failed to clone from github"
        exit 1
    }
```

sources/functions/qbittorrent
flying-sausages: @liaralabs not sure why this was marked as resolved, you can get rid of the `cd /tmp`. If you need to be in the same dir for `./configure`, might as well ` cd /tmp/qbitttorrent` just before it. None of the paths between L121 and L 126 need to be based on cwd from what I can see.

sources/functions/utils
```diff
@@ -174,19 +174,23 @@ function check_ram() {
...
-        echo_error "Cannot continue with install because swap file could not be created. Does device have 2GB of free space?"
-        exit 1
-    }
-    mkswap /.swapfile > /dev/null 2>&1
-    swapon /.swapfile > /dev/null 2>&1
-    echo_progress_done 'Swap enabled'
+function check_swap_on() {
+    if ! check_ram 2G; then
+        echo_progress_start "Setup is enabling the use of a temporary 2GB swap file"
+        dd if=/dev/zero of=/.swapfile bs=1M count=2048 > /dev/null 2>&1 || {
```
flying-sausages, line 180: Do we want to check if this is available to begin with? if <2GB RAM is the case, chance is the device is running on like minimal storage mounted on root too

sources/functions/qbittorrent
```diff
@@ -41,48 +36,110 @@ function whiptail_qbittorrent() {
...
+    git clone -b ${libtorrent_branch} https://github.com/arvidn/libtorrent >> ${log} 2>&1
+    cd /tmp/libtorrent
+    if [[ -f /root/libtorrent-${libtorrent_branch}.patch ]]; then
+        patch -p1 < /root/libtorrent-${libtorrent_branch}.patch >> ${log} 2>&1
+    fi
```
flying-sausages, line 98: Might want to add some echo here about "Found a patch, applying"
```suggestion
    if [[ -f /root/libtorrent-${libtorrent_branch}.patch ]]; then
        patch -p1 < /root/libtorrent-${libtorrent_branch}.patch >> ${log} 2>&1 || {
            echo _error "Something went wrong when patching"
            exit 1
        }
        echo_info "Found libtorrent patch and applied it"
    fi
```

flying-sausages:
If you give me some specific scenarios I should test to verify this works, I'll do it this weekend.

sources/functions/deluge
```diff
@@ -49,114 +51,204 @@ function remove_ltcheckinstall() {
...
+    cd -
+    cd /tmp/libtorrent/bindings/python
+    /opt/boost_${BOOST_VERSION}/b2 -j$(nproc) python=${pythonmajor} crypto=openssl variant=release libtorrent-link=static boost-link=static install_module python-install-path=/tmp/dist/libtorrent-python/usr/local/lib/python${pythonmajor}/dist-packages >> ${log} 2>&1
+    tag=" static with c++14 march=native"
+    mkdir -p /root/dist
+    fpm -f -C /tmp/dist/libtorrent-python -p /root/dist/${python}-libtorrent_VERSION.deb -s dir -t deb -n ${python}-libtorrent --version ${VERSION} --description "Libtorrent rasterbar python binding compiled by swizzin$tag" >> ${log} 2>&1 || {
+        echo_error "Error packaging ${python}-libtorrent. Setup will exit"
+        exit 1
+    }
+    cd -
```
flying-sausages, line 128: What are these `cd -` being used for?

sources/functions/qbittorrent
liaralabs: You have suggested changes to outdated code

sources/functions/qbittorrent
liaralabs: I believe you are commenting on outdated code here again.

sources/functions/deluge
liaralabs: `cd -` returns to the previous working directory.

This must be done as you cannot `rm` a folder while located inside of it, thus we must move back to the previous position. Previously, `/tmp` was the directory used for this.

sources/functions/qbittorrent
flying-sausages: Yikes, my bad!

sources/functions/deluge
flying-sausages: What about `cd`ing to `/` instead? making it relative through the use of `-` might be handy but things could go wrong if the previous location no longer exists for whatever reason

sources/functions/qbittorrent
flying-sausages: Yup, looks like it. And it looks like you already thought of the exact same solution i had in mind ha!

sources/functions/deluge
liaralabs: What was wrong with `/tmp` then?

flying-sausages: Now that I know this is dependent on these cds, probably nothing. Not sure if it was to temp or a die under it. Sorry about that.

PR #599, opened 2021-03-21
feat(Requestrr): Add Requestrr

brettpetch:
## Description
Add Requestrr to box installs

## Fixes issues: 
Fixes no issues.

## Proposed Changes:
- Add Requestrr to the stack

## Change Categories

- New feature / Application 

## Checklist

    - PR link: https://github.com/liaralabs/docs.swizzin.ltd/pull/56
    - PR link: https://github.com/liaralabs/swizzin_dashboard/pull/32

## Test scenarios

Has been tested on AMD64, should also be respectful of ARM.

### Architectures

| Focal 		|			|			|	✅		|				|
| Bionic		|			|			|			|				|
| Buster		|	✅		|			|			|				|
| Stretch		|			|			|			|				|
| Raspbian  	|			|			|			|				|

### 🛠🛠 TODO
- [x] Docs (@HarleyL7)
- [x] Updater script
- [x] Removal script
- [x] Panel stuffs
- [x] Testing
- [x] Auto-config?
- [ ] Unattended login with swizz credentials

### ❌❌ Currently failing
Tested. Nginx + Installs + removals succeed.

Continues to serve on all interfaces when nginx is enabled, although 400's on all requests to the port; except those via the reverse-proxy.


brettpetch:
Ok, almost all is done. Only thing I have yet to do are docs and testing. 

Solution for the exposed port: 4545, what if we were to modify iptables; just a suggestion, but not sure if that's something we've ever done before in the project. 

/b

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,178 @@
...
+. /etc/swizzin/sources/functions/os
+
+user=$(cut -d: -f1 < /root/.master.info)
+
+echo_progress_start "Downloading source files"
+case "$(_os_arch)" in
+    "amd64") wget -qO "/tmp/requestrr.zip" "$(curl -sNL https://api.github.com/repos/darkalfx/requestrr/releases/latest | grep -Po 'ht(.*)linux-x64(.*)zip')" >> ${log} 2>&1 ;;
+    "armhf") wget -qO "/tmp/requestrr.zip" "$(curl -sNL https://api.github.com/repos/darkalfx/requestrr/releases/latest | grep -Po 'ht(.*)linux-arm(.*)zip')" >> ${log} 2>&1 ;;
+    "arm64") wget -qO "/tmp/requestrr.zip" "$(curl -sNL https://api.github.com/repos/darkalfx/requeAstrr/releases/latest | grep -Po 'ht(.*)linux-arm64(.*)zip')" >> ${log} 2>&1 ;;
+    *)
```
flying-sausages, line 21: I'd set the variable in the switch case and then do the wget outside so we can handle a failure there

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,178 @@
...
+echo_progress_start "Extracting archive"
+unzip /tmp/requestrr.zip -d /opt/ >> "$log" 2>&1
+rm /tmp/requestrr.zip
+mv requestrr* requestrr
+echo_progress_done "Archive extracted"
+
+touch /install/.requestrr.lock
+
+echo_progress_start "Creating requestrr user and setting permssions"
+useradd -M --shell=/bin/false requestrr
```
flying-sausages, line 37: why not use `--system` and set the homedir to the folder in opt? not sure what `-M`

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,178 @@
...
+[Unit]
+Description=Requestrr Daemon
+After=syslog.target network.target
+
+[Service]
+User=requestrr
+Type=simple
+WorkingDirectory=/opt/requestrr/
+ExecStart=/opt/requestrr/Requestrr.WebApi
+TimeoutStopSec=20
```
flying-sausages, line 155: Is this necessary? haven't seen this used a lot

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,178 @@
...
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+if [[ -f /install/.nginx.lock ]]; then
+    echo_progress_start "Installing nginx configuration"
+    #TODO what is this sleep here for? See if this can be fixed by doing a check for whatever it needs to
+    sleep 10
```
flying-sausages, line 166: Does this need to be here? sounds like something i wrote that made it here hahahaahah

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,178 @@
...
+    bash /usr/local/bin/swizzin/nginx/requestrr.sh
+    systemctl daemon-reload
```
flying-sausages, line 168: why's there a daemon reload here? if the nginx scriptis modifying the service then that should be there instead

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,178 @@
...
+    systemctl -q reload nginx
+    echo_progress_done "Nginx configured"
+else
+    echo_info "Requestrr will be available on port 4545. Secure your installation manually through the web interface."
+fi
+
+systemctl -q daemon-reload
+systemctl enable --now -q requestrr
+sleep 1
```
flying-sausages, line 177: no need

scripts/install/requestrr.sh
flying-sausages: looks like it's not being modified, no need for it at all i think

flying-sausages (changes requested):
Still gotta test this

scripts/install/requestrr.sh
brettpetch: this was the dev's suggestion in the docs, is there a better way to implement?

scripts/install/requestrr.sh
brettpetch: will remove shortly.

scripts/install/requestrr.sh
brettpetch: will be changed to ``useradd --system -d /opt/requestrr requestrr``

scripts/install/requestrr.sh
brettpetch: removed.

scripts/install/requestrr.sh
brettpetch: removed the daemon-reload and sleep 10

brettpetch:
going to go run a test on raspi.

brettpetch:
Testing succeeds.

brettpetch:
I think @HarleyL7 is going to write docs tonight from what I've gathered. I will update the PR after he has finished.

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,178 @@
...
+if ! curl "$dlurl" -L -o /tmp/requestrr.zip >> "$log" 2>&1; then
+    echo_error "Download failed, exiting"
+    exit 1
+fi
+echo_progress_done "Source downloaded"
+
+echo_progress_start "Extracting archive"
+unzip -q /tmp/requestrr.zip -d /opt/ >> "$log" 2>&1
+rm /tmp/requestrr.zip
+mv /opt/requestrr* /opt/requestrr
```
flying-sausages, line 34: If you get the tar, you can just extract it with strip-component flag to get rid of the first dir. This could get hacky and wonky

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,178 @@
...
+echo_progress_done "Archive extracted"
+
+touch /install/.requestrr.lock
+
+echo_progress_start "Creating requestrr user and setting permssions"
+useradd --system -d /opt/requestrr/ requestrr
+chown -R requestrr:requestrr /opt/requestrr/
+chmod +x /opt/requestrr/Requestrr.WebApi
```
flying-sausages, line 42: mmmmmm maybe owner only?
```suggestion
chmod u+x /opt/requestrr/Requestrr.WebApi
```

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,178 @@
...
+      "SearchNewRequests": true,
+      "MonitorNewRequests": true,
+      "UseSSL": false,
+      "Version": "2"
+    },
+    "Sonarr": {
+      "Hostname": "",
+      "Port": 8989,
+      "ApiKey": "",
+      "BaseUrl": "",
```
flying-sausages, line 109: we _could_ fill these in, seeing as they are all in the config xml file in ~/.config. Would be slick. Could help out with that.

scripts/nginx/requestrr.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+# Nginx Configuration for requestrr
+master=$(cut -d: -f1 < /root/.master.info)
```
flying-sausages, line 3: We actually have a function for this now which should be available without having to source it, should be mentioned in `sources/globals.sh`

scripts/nginx/requestrr.sh
```diff
@@ -0,0 +1,29 @@
...
+  ### This shouldn't be needed
+  ### auth_basic "What's the password?";
+  ### auth_basic_user_file /etc/htpasswd.d/htpasswd.${master};
+}
+SRC
+    cat > /opt/requestrr/appsettings.json << SET
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Warning"
```
flying-sausages, line 23: Is this required? Just hoping it won't crash and fill the disk when on some SD card based system or sth

scripts/update/requestrr.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
flying-sausages, line 1: If the update and the install share large portions of code, I would encourage you to abstract them into a function file you can then source in both the installer and the updater. Less code duplication, one place to keep track of it.

scripts/update/requestrr.sh
```diff
@@ -0,0 +1,41 @@
...
+. /etc/swizzin/sources/functions/os
```
flying-sausages, line 2: This should be sourced in `globals.sh` for you already iirc

scripts/update/requestrr.sh
```diff
@@ -0,0 +1,41 @@
...
+        *)
+            echo_error "Arch not supported"
+            exit 1
+            ;;
+    esac
+
+    if ! curl "$dlurl" -L -o /tmp/requestrr.zip >> "$log" 2>&1; then
+        echo_error "Download failed, exiting"
+        exit 1
+    fi
```
flying-sausages, line 19: just a stylistic thing, but we're using ` || {}` for this

```suggestion
    curl "$dlurl" -L -o /tmp/requestrr.zip >> "$log" 2>&1 || {
        echo_error "Download failed, exiting"
        exit 1
    }
```
looks slicker imo

scripts/update/requestrr.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo_progress_done "Source downloaded"
+    echo_progress_start "Extracing and overwriting (no configs will be lost)"
```
flying-sausages, line 21:
```suggestion
    echo_progress_start "Replacing source code"
```

scripts/update/requestrr.sh
```diff
@@ -0,0 +1,41 @@
...
+    if [[ -f /install/.nginx.conf ]]; then
+        bash /usr/local/bin/swizzin/nginx/requestrr.sh
+        systemctl -q reload nginx
+    fi
+    echo_progress_done "Config patched"
+
+    echo_progress_start "Restarting services"
+    systemctl daemon-reload
+    systemctl -q enable --now requestrr
+    echo_progress_done "Requestrr has been updated."
```
flying-sausages, line 40:
```suggestion
    echo_progress_done "Requestrr has been updated."
```

scripts/update/requestrr.sh
flying-sausages, line 1: Please move the file to the upgrade folder https://swizzin.ltd/dev/structure

flying-sausages (commented):
Picking some smaller nits, once these are resolved I'd be happy with merging (after I test it for myself)

scripts/update/requestrr.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+. /etc/swizzin/sources/functions/os
+if [[ -f /install/.requestrr.lock ]]; then
```
flying-sausages, line 3: I'd negate this and kill the script if that's not present, seeing as this should be an upgrader. I don't see the other case handled as is.

scripts/update/requestrr.sh
```diff
@@ -0,0 +1,41 @@
...
+    systemctl disable --now requestrr
```
flying-sausages, line 4:
```suggestion
    echo_info "Upgrading Requestrr"
    systemctl disable --now requestrr
```

scripts/install/requestrr.sh
brettpetch: I didn’t see the dev has a tar available. Seems like zips only from the [Releases Page](https://github.com/darkalfx/requestrr/releases/tag/V1.0.17).

Originally I did want to go and use a tar for this reason. Unfortunately it doesn’t seem to be an option at this time.

scripts/install/requestrr.sh
brettpetch: I wasn’t sure how this password was being salted and hashed. It didn’t look like htpasswd; but I’ll go for a look see and do some playing with it this afternoon.

scripts/nginx/requestrr.sh
brettpetch: I've set to none. will see how it goes.

scripts/install/requestrr.sh
brettpetch: Going to def. play with this in a bit -- going for a walk, then will get started on it.

scripts/update/requestrr.sh
brettpetch: did this :)

scripts/install/requestrr.sh
brettpetch: Have consulted with a few peeps and they made the argument that this should be opt-in -- what says you?

flying-sausages: Nah make it automatic if it's installed, whoever would want to opt out will know what they need to do. They'll have to change things anyway. 80% usecase says opt-in.

brettpetch: @flying-sausages, what if I do an {ask} y/n?; or would this screw up on an unattended install...

/b

flying-sausages: nah let's just do it default lol. opt-out through environment variable _maybe_. Bazarr does the same thing. Ombi should too, and we should do the same thing with overseerr whenever it's possible.

brettpetch: Alrighty. I'll merge in the stuffs and do some testing 😜

flying-sausages:
I mentioned this before, but I do not know how to functionally test this. I don't really have a personal discord server or something. Is there any easy path to check this works right?

brettpetch:
> 
> 
> I mentioned this before, but I do not know how to functionally test this. I don't really have a personal discord server or something. Is there any easy path to check this works right?

I've tested that... I just need to add liara's snippet to make user credentials work by default.

brettpetch:
I have a discord if you'd like to try it out @flying-sausages... Just let me know and I'll get one setup for us.

flying-sausages:
> I've tested that...

I'm sure you did, but I'd like to test it out myself before approving this ha




> I have a discord if you'd like to try it out @flying-sausages... Just let me know and I'll get one setup for us.

sure why not

brettpetch:
going to the farm but should be back in hr. will send you creds and junk in a bit.

/b

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,182 @@
...
+_get_sonarr_vars
+_get_radarr_vars
+echo_progress_done "Apps have been added and will be added to Requestrr config."
+
+cat > /opt/requestrr/SettingsTemplate.json << CFG
+{
+  "Authentication": {
+    "Username": "",
+    "Password": "",
+    "PrivateKey": "[PRIVATEKEY]"
```
flying-sausages, line 44: Does this get replaced after install?

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,182 @@
...
+echo_progress_start "Grabbing list of apps and integrating with Requestrr"
+_get_sonarr_vars
+_get_radarr_vars
+echo_progress_done "Apps have been added and will be added to Requestrr config."
+
+cat > /opt/requestrr/SettingsTemplate.json << CFG
+{
+  "Authentication": {
+    "Username": "",
+    "Password": "",
```
flying-sausages, line 43: 🐍 🧂

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,182 @@
...
+  },
+  "DownloadClients": {
+    "Ombi": {
+      "Hostname": "",
+      "Port": 3579,
+      "ApiKey": "",
+      "ApiUsername": "",
+      "BaseUrl": "",
+      "UseSSL": false,
+      "Version": "3"
```
flying-sausages, line 70: should this be `4`?

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,182 @@
...
+      "ApiKey": "",
+      "DefaultApiUserID": "",
+      "UseSSL": false,
+      "Version": "1"
+    },
+    "Radarr": {
+      "Hostname": "${r_address}",
+      "Port": ${r_port},
+      "ApiKey": "${r_key}",
+      "BaseUrl": "/${r_base}",
```
flying-sausages, line 84: This should just be standard
```suggestion
      "BaseUrl": "/radarr",
```

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,182 @@
...
+    "Overseerr": {
+      "Hostname": "",
+      "Port": 5055,
+      "ApiKey": "",
+      "DefaultApiUserID": "",
+      "UseSSL": false,
+      "Version": "1"
+    },
+    "Radarr": {
+      "Hostname": "${r_address}",
```
flying-sausages, line 81:
```suggestion
      "Hostname": "127.0.0.1",
```

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,182 @@
...
+    "Command": "movie"
+  },
+  "TvShows": {
+    "Client": "Disabled",
+    "Command": "tv",
+    "Restrictions": "None"
+  },
+  "Port": 4545,
+  "BaseUrl" : "/requestrr",
+  "Version": "1.12.0"
```
flying-sausages, line 134: Is this still actual?

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,182 @@
...
+echo_progress_done "Requestrr user has been created & permissions set."
+
+echo_progress_start "Applying Requestrr config..."
+
+echo_progress_start "Grabbing list of apps and integrating with Requestrr"
+_get_sonarr_vars
+_get_radarr_vars
+echo_progress_done "Apps have been added and will be added to Requestrr config."
+
+cat > /opt/requestrr/SettingsTemplate.json << CFG
```
flying-sausages, line 39: This thing is probably gonna change format over time. I think `cat`ing this hardcoded version could prove not very maintainable.

How about we try this `jq` trick? Not sure this one will execute, but it's like `sed -i` but for JSON. There are also alternatives with `sponge` but I'm not sure we need the extra dependency for this.
```bash
cat <<< $(jq '.Authentication.Username = "abcde"' SettingsTemplate.json) > SettingsTemplate.json
```
`jq` is a bit specific with the way it works with bash variables but there are tricks I did for this in the transmission functions already

flying-sausages: This way we could just keep it to the things we want to change, and not hardcode the whole file here

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,182 @@
...
+
+if [[ -f /install/.nginx.lock ]]; then
+    echo_progress_start "Installing nginx configuration"
+    bash /usr/local/bin/swizzin/nginx/requestrr.sh
+    systemctl -q reload nginx
+    echo_progress_done "Nginx configured"
+else
+    echo_info "Requestrr will be available on port 4545. Secure your installation manually through the web interface."
+    cat > /opt/requestrr/appsettings.json << SET
+{
```
flying-sausages, line 167: could also do this jq trick here if it's just for allowed hosts

scripts/nginx/requestrr.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# Nginx Configuration for requestrr
+#shellcheck source=sources/functions/users
+. /etc/swizzin/sources/functions/users
+master=$(_get_master_username)
```
flying-sausages, line 5:
```suggestion
master="$(_get_master_username)"
```

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,182 @@
...
+WantedBy=multi-user.target
+EOF
+
+if [[ -f /install/.nginx.lock ]]; then
+    echo_progress_start "Installing nginx configuration"
+    bash /usr/local/bin/swizzin/nginx/requestrr.sh
+    systemctl -q reload nginx
+    echo_progress_done "Nginx configured"
+else
+    echo_info "Requestrr will be available on port 4545. Secure your installation manually through the web interface."
```
flying-sausages, line 165: If the username and the salted password hash is in the config, does this still need to be done manually?

scripts/nginx/requestrr.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# Nginx Configuration for requestrr
+#shellcheck source=sources/functions/users
+. /etc/swizzin/sources/functions/users
```
flying-sausages, line 4: already in `glbals.sh`
```suggestion
```

scripts/nginx/requestrr.sh
```diff
@@ -0,0 +1,31 @@
...
+location ^~ /requestrr {
+  proxy_pass        http://127.0.0.1:4545/requestrr;
+  proxy_set_header Host \$proxy_host;
+  proxy_set_header X-Forwarded-For \$proxy_add_x_forwarded_for;
+  proxy_set_header X-Forwarded-Proto \$scheme;
+  proxy_redirect off;
+  # Basic Auth if Wanted
+  ### This shouldn't be needed
+  ### auth_basic "What's the password?";
+  ### auth_basic_user_file /etc/htpasswd.d/htpasswd.${master};
```
flying-sausages, line 18: I would actually say that if there's only one account that can use this, might as well not set up app-side auth for the nginx-served app. Just like sonarr, just do basic auth and unprotect the app.

scripts/upgrade/requestrr.sh
```diff
@@ -0,0 +1,41 @@
...
+echo_progress_start "Replacing source code"
+
+unzip -q /tmp/requestrr.zip -d /tmp/ >> "$log" 2>&1
+rm /tmp/requestrr.zip
+cp -RT /tmp/requestrr*/ /opt/requestrr
+chown -R requestrr:requestrr /opt/requestrr
+rm -rf /tmp/requestrr*
+echo_progress_done "Extracted and overwrote existing files."
+
+echo_progress_start "Patching config"
```
flying-sausages, line 19: Why does this need to be done during an upgrade?

scripts/upgrade/requestrr.sh
```diff
@@ -0,0 +1,41 @@
...
+if [[ -f /install/.nginx.conf ]]; then
+    bash /usr/local/bin/swizzin/nginx/requestrr.sh
```
flying-sausages, line 21: this will not replace the nginx conf seeing as that script will only put it there it in case it does not exist, FYI

flying-sausages (commented):
☠️ ⛏️

brettpetch:
Python bits for generating hashes + inserting password: I'll do this tonight when I've got a sec... have to see where else uses python and see how to do this properly.

```python
def set_access_details(password):
    import uuid
    import base64
    import hashlib
    salt = uuid.uuid4()
    salt_bytes = salt.bytes
    password = str.encode(password)
    hashed_password = hashlib.pbkdf2_hmac('sha1', password, salt_bytes, 10000, dklen=32)
    b64_password = base64.b64encode(hashed_password).decode("utf-8")
    return b64_password, salt
```

scripts/install/requestrr.sh
brettpetch: this will be updated once I've committed it.

sources/functions/appconfigs
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
brettpetch: there's actually a better way to do this without dependencies that I've found. Will update.

sources/functions/appconfigs
```diff
@@ -0,0 +1,33 @@
...
+#shellcheck source=sources/functions/users
+. /etc/swizzin/sources/functions/users
+
+apt_install libxml2-utils
```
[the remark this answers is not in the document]
brettpetch: I found out that we already have a way to cherrypick xml with sed.

scripts/upgrade/requestrr.sh
brettpetch: I'll go for another look and see.

scripts/nginx/requestrr.sh
brettpetch: The app enforces it's own protection.

scripts/install/requestrr.sh
```diff
@@ -0,0 +1,182 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
brettpetch: yeah, sure.

scripts/upgrade/requestrr.sh
brettpetch: lmao ``/install/.nginx.conf`` never exists... i must've written this when I was dumb.

liaralabs:
Just a heads up that it looks like this PR is going to need to be rebased in order to be considered for merging.

With the number of commits in this PR, it may be easier to simply start over with a new branch/PR as it looks like a rebase was done partway through the process.

flying-sausages:
Closing in favour of #718 seeing as mr goose is a member

PR #615, opened 2021-04-05
feature: Authelia SSO wip

userdocs:
## Description

Authelia single sign on solution to replace the basic auth method used.

## Fixes issues: 
- na

## Proposed Changes:

- TODO - nginx should be installed by default
- TODO - all apps with basic auth changed to use authelia
- TODO - exception for api urls with token based auth via the authelia config

## Change Categories
- Bug fix 
- New feature 

## Checklist

    - PR link: 
    - PR link: 
- [ ] Code is formatted [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)
- [ ] Shellcheck isn't screaming [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)

### Architectures

| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|


flying-sausages:
> TODO - All logins will be handled by Authelia To do this Authelia needs to take control of the / location.

So this is a hard dependency on the BaseURL support in panel? https://github.com/liaralabs/swizzin_dashboard/issues/25

flying-sausages:
Would be nice to support also alternative BaseURLs for authelia so that I can drop it someplace else than `/` but maybe `/login`, but i have no idea whether that is within the spec of what authelia can do.

We can store values in our "swis db" so that you don't have to always run some grep/awk type data retrieval, I assume that would make things easier for you a bit
https://github.com/swizzin/swizzin/blob/master/sources/functions/swizdb
The documentation for that should be self-sufficient :)

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+# Create the loction for the stored binary
+mkdir -p "/opt/authelia"
+# Download the binary
+wget -qO "/opt/authelia/authelia-linux-amd64.tar.gz" "${authelia_url}"
+# Extract the specific file we need and nothing else.
+tar -xf "/opt/authelia/authelia-linux-amd64.tar.gz" -C "/opt/authelia/" 'authelia-linux-amd64'
+# Symlink the extracted binary authelia-linux-amd64 to authelia
+ln -fsn "/opt/authelia/authelia-linux-amd64" "/opt/authelia/authelia"
+# Remove the archive we no longer need
+[[ -f "/opt/authelia/authelia-linux-amd64.tar.gz" ]] && rm -f "/opt/authelia/authelia-linux-amd64.tar.gz"
```
flying-sausages, line 34: This whole section needs to be arch agnostic. The binary downloaded already is but the filenames are hardcoded, and the file on L30 needs to be retrieved dynamically especially i suppose

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+. /etc/swizzin/sources/functions/utils
+. /etc/swizzin/sources/functions/os
+
+# Get the current version using a git ls-remote tag check
+authelia_latestv="$(git ls-remote -t --refs https://github.com/authelia/authelia.git | awk '{sub("refs/tags/", "");sub("(.*)-alpha(.*)", ""); print $2 }' | awk '!/^$/' | sort -rV | head -n1)"
+# Create the download url using the version provided by authelia_latestv
+case "$(_os_arch)" in
+    "amd64") authelia_arch="amd64" ;;
+    "arm32") authelia_arch="arm32v7" ;;
+    "arm64") authelia_arch="arm64v8" ;;
```
flying-sausages, line 17:
```suggestion
    "armhf") authelia_arch="arm32v7" ;;
    "arm64") authelia_arch="arm64v8" ;;
```

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+port: 9091
+
+server:
+  read_buffer_size: 4096
+  write_buffer_size: 4096
+  path: "authelia"
+
+theme: dark
+log_level: debug
+log_file_path: /etc/authelia/authelia.log
```
flying-sausages, line 59: not `/var/log/`?

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+host: 127.0.0.1
+port: 9091
+
+server:
+  read_buffer_size: 4096
+  write_buffer_size: 4096
+  path: "authelia"
+
+theme: dark
+log_level: debug
```
flying-sausages, line 58: Will this be necessary? and on the flipside, can we use this to get fail2ban on this? Would be nice to automatically supply that conf

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+      memory: 1024
+      parallelism: 8
+
+access_control:
+  default_policy: deny
+  rules:
+    - domain: ${ex_ip}
+      resources:
+        - "^/(sonarr|radarr|jackett)/api.*$"
+      policy: bypass
```
flying-sausages, line 83: Just thinking towards the future, should this section come pre-filled with all possible API endpoints you can install in swizzin, or should each app installer have its own "authelia hook" which would add it here and then reload authelia?

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+
+    - domain: ${ex_ip}
+      policy: one_factor
```
flying-sausages, line 86: Should we have a thinker about the policy to implement by default?

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+
+session:
+  name: authelia_session
+  secret: ${insecure_secret}
+  expiration: 1h
+  inactivity: 5m
+  remember_me_duration: 1M
```
flying-sausages, line 93: Should we discuss these values? 5 minutes for inactivity seems rather short

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+  domain: ${ex_ip}
+
+regulation:
+  max_retries: 3
```
flying-sausages, line 97: 5 pls :'(

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+# This file can be used if you do not have an LDAP set up.
+
+users:
+  username:
+    displayname: "username"
+    password: "${password_hash}"
+    email: username@swizzin.com
+    groups:
+      - admins
+      - dev
```
flying-sausages, line 126: Should username here be replaced with the master uname? or is this a template?

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+Restart=always
+RestartSec=2
+TimeoutStopSec=5
+SyslogIdentifier=authelia
+
+[Install]
+WantedBy=default.target
+AUTHELIA_SERVICE
+# Symlink our service file from /etc/authelia/authelia.service to /etc/systemd/system/authelia.service
+ln -fsn /etc/authelia/authelia.service /etc/systemd/system/authelia.service
```
flying-sausages, line 146: huh interesting, why not cat it directly there?

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+
+authelia_url="https://github.com/authelia/authelia/releases/download/${authelia_latestv}/authelia-linux-${authelia_arch}.tar.gz"
+# Create the loction for the stored binary
+mkdir -p "/opt/authelia"
+# Download the binary
+wget -qO "/opt/authelia/authelia-linux-amd64.tar.gz" "${authelia_url}"
+# Extract the specific file we need and nothing else.
+tar -xf "/opt/authelia/authelia-linux-amd64.tar.gz" -C "/opt/authelia/" 'authelia-linux-amd64'
+# Symlink the extracted binary authelia-linux-amd64 to authelia
+ln -fsn "/opt/authelia/authelia-linux-amd64" "/opt/authelia/authelia"
```
flying-sausages, line 32: curious, why symlink this? I mean, it will work, just wondering why tho

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+# Symlink our service file from /etc/authelia/authelia.service to /etc/systemd/system/authelia.service
+ln -fsn /etc/authelia/authelia.service /etc/systemd/system/authelia.service
+# Install nginx stuff
+if [[ -f /install/.nginx.lock ]]; then
+    echo_progress_start "Installing nginx config"
+    bash "/usr/local/bin/swizzin/nginx/authelia.sh"
+    systemctl reload nginx
+    echo_progress_done "Nginx config installed"
+fi
+# enabel the service
```
flying-sausages, line 154:
```suggestion
# enable the service
```

flying-sausages: ;)

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
+#!/bin/bash
+#
```
flying-sausages, line 2: This whole file should be re-written with segments of code separated into functions.

That way we can also make a simple implementation for MU-later, which will only call some portions of this when the install script is being invoked with an argument which will hold the name of the new user being added

scripts/install/authelia.sh
flying-sausages: What about all the other users?

scripts/nginx/authelia.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
+#
+# Get our externial IP
+ex_ip="$(ip -br a | sed -n 2p | awk '{ print $3 }' | cut -f1 -d'/')"
```
flying-sausages, line 4: Maybe I missed this but I don't see this variable used here. Also, if it is, I'm curious how this would work for our non-static-IP friends

scripts/nginx/authelia.sh
```diff
@@ -0,0 +1,120 @@
...
+# Create an /etc/nginx/apps subdirectory we need
+mkdir -p "/etc/nginx/apps/authelia"
+# Fix nginx to not try to load a directory when using an include
+sed 's|include /etc/nginx/apps/\*;|include /etc/nginx/apps/*.conf;|g' -i /etc/nginx/sites-enabled/default
+# Create the main reverse proxy - it contains 2 location driectoive for this example.
+cat > "/etc/nginx/apps/authelia.conf" << AUTHELIA_NGINX_1
+set \$upstream_authelia http://127.0.0.1:9091;
```
flying-sausages, line 11: 9091 is already the default RPC port for transmission FYI. Every nth user users has port 9091+n-1, so you might want to go under 9091 maybe.

scripts/nginx/authelia.sh
```diff
@@ -0,0 +1,120 @@
...
+
+# Virtual endpoint created by nginx to forward auth requests.
+location /authelia/api/verify {
+    internal;
+    proxy_pass_request_body off;
+    proxy_pass \$upstream_authelia\$auth_type;
+    proxy_set_header Content-Length "";
+
+    # Timeout if the real server is dead
+    proxy_next_upstream error timeout invalid_header http_500 http_502 http_503;
```
flying-sausages, line 28: 504s?

scripts/remove/authelia.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+#
+# |& was added to Bash 4 as an abbreviation for 2>&1 |. - https://tldp.org/LDP/abs/html/io-redirection.html
```
flying-sausages, line 3: mmmmmm are we sure everyone is on bash4 actually? something something debian

scripts/remove/authelia.sh
```diff
@@ -0,0 +1,23 @@
...
+#
+if [[ ! -f /install/.authelia.lock ]]; then
+    echo_warn "Authelia is not installed"
+    exit 1
+else
+    systemctl disable -q --now authelia |& tee -a "${log}" &> /dev/null
+    rm -rf /etc/authelia
```
flying-sausages, line 10: don't forget to remove the symlink in /etc/systemd/system as that will remain there and be all broken and sad

scripts/upgrade/authelia.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+#
+if [[ ! -f /install/.authelia.lock ]]; then
+    echo_warn "Authelia is not installed"
+    exit 1
+else
+    echo_progress_start "Updating Authelia"
+    systemctl stop -q authelia |& tee -a $log
+    # Set the current version using a git ls-remote tag check
+    authelia_latestv="$(git ls-remote -t --sort=-v:refname --refs https://github.com/authelia/authelia.git | awk '{sub("refs/tags/", "");sub("(.*)-alpha(.*)", ""); print $2 }' | head -n1)"
```
flying-sausages, line 10: Any chance we can check if the upgrade is necessary or not?

scripts/upgrade/authelia.sh
```diff
@@ -0,0 +1,33 @@
...
+    # Create the download url using the version provided by authelia_latestv
+    authelia_url="https://github.com/authelia/authelia/releases/download/${authelia_latestv}/authelia-linux-amd64.tar.gz"
```
flying-sausages, line 12: This one is mot multi-arch

scripts/install/authelia.sh
userdocs: it can be done here on using nginx locations.

scripts/install/authelia.sh
userdocs: this is just the defaults.

scripts/nginx/authelia.sh
userdocs: these are the defaults.

scripts/remove/authelia.sh
userdocs: something something google it.

https://packages.debian.org/jessie/bash

scripts/install/authelia.sh
```diff
@@ -0,0 +1,159 @@
...
+#                         Users Database                      #
+###############################################################
+
+# This file can be used if you do not have an LDAP set up.
+
+users:
+  username:
+    displayname: "${username}"
+    password: "${password_hash}"
+    email: ${username}@swizzin.com
```
flying-sausages, line 123:
```suggestion
    email: ${username}@localhost
```

scripts/upgrade/authelia.sh
userdocs: how else will it get updated? i could be automated like this.

```bash
if [[ "$authelia_local_version" != "$authelia_remote_version" ]]; then
	wget -qO "$HOME/authelia.tar.gz" "https://github.com/authelia/authelia/releases/download/v${authelia_remote_version}/authelia-linux-amd64.tar.gz"
	tar -xf "$HOME/authelia.tar.gz" -C "$HOME/bin/" authelia-linux-amd64
	mv -f "$HOME/bin/authelia-linux-amd64" "$HOME/bin/authelia"
	[[ -f "$HOME/authelia.tar.gz" ]] && rm -f "$HOME/authelia.tar.gz"
	systemctl --user restart authelia.service
	echo "Authelia updated and restarted"
fi
```

flying-sausages: yeah basically that, would be sweet

scripts/install/authelia.sh
userdocs: because they decided to include the arch in the binary name. authelia-linux-ARCH. Since I cannot rename while extracting it's less complex to just create a symlink instead trying to rm, mv or cp, rm.

scripts/install/authelia.sh
```diff
@@ -0,0 +1,163 @@
...
+#
+authelia_url="https://github.com/authelia/authelia/releases/download/${authelia_latestv}/authelia-linux-${authelia_arch}.tar.gz"
+# Create the loction for the stored binary
+mkdir -p "/opt/authelia"
+# Download the binary
+wget -qO "/opt/authelia/authelia-linux-${authelia_arch}.tar.gz" "${authelia_url}"
+# Extract the specific file we need and nothing else.
+tar -xf "/opt/authelia/authelia-linux-${authelia_arch}.tar.gz" -C "/opt/authelia/" "authelia-linux-${authelia_arch}"
+# Symlink the extracted binary authelia-linux-${authelia_arch} to authelia
+ln -fsn "/opt/authelia/authelia-linux-${authelia_arch}" "/opt/authelia/authelia"
```
flying-sausages, line 32:
```suggestion
mv "/opt/authelia/authelia-linux-${authelia_arch}" "/opt/authelia/authelia"
```

PR #616, opened 2021-04-06
fix: make sure command is executed before nginx is reloaded.

userdocs:
should have put this change before the nginx reload command or it won't do anything helpful.


flying-sausages (approved):

flying-sausages:
yeah let's not go and update an updater lol

PR #621, opened 2021-04-09
feature: function : app ports

userdocs:
## Description

This function uses port values stored in a bash associative array, like this:

```bash
app_port[authelia]=9091
app_port[filebrowser]=9090
app_port[jackett]=9117
```

And calls the function passing the script name as the positional parameter using this command

```bash
_get_app_port "$(basename -- "$0" \.sh)"
```
So if the `jackett.sh` script uses this command:

```bash
app_proxy_port="$(_get_app_port "$(basename -- "$0" \.sh)")"
```

`app_proxy_port` will be set to `9117`

Since all scripts are named as their respective application defining the port as the script name, minus the extension, will give a predictable and constant result.

**Other considerations:**

- The array does not have to be stored in the function.
- The errors could be mad set to use existing functions or provide a different result.

## Fixes issues: 
- na
- 
## Proposed Changes:
All scripts use this command and variable instead of hardcoding the port

```bash
app_proxy_port="$(_get_app_port "$(basename -- "$0" \.sh)")"
```

## Change Categories
- New feature 
- Change in `functions` 
- Contribution flow 

## Checklist

    - PR link: 
    - PR link: 

### Architectures

| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|

### DIFF
--- a/sources/functions/app_port
+++ b/sources/functions/app_port
@@ -7,7 +7,7 @@ app_port[authelia]=9091
 app_port[filebrowser]=9090
 app_port[jackett]=9117
 #
-# Use this function to get a port from the array using the command inside an install script app_port="$(_get_app_port "$(basename -- "$0")")"
+# Use this function to get a port from the array using the command inside an install script app_proxy_port="$(_get_app_port "$(basename -- "$0" \.sh)")"
 _get_app_port() {
     if [[ -n "${1}" ]]; then
         if [[ -n ${app_port[$1]+$1} ]]; then
--- a/sources/functions/app_port
+++ b/sources/functions/app_port
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+#
+declare -A app_port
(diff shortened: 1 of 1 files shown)


sources/functions/app_port
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+#
+declare -A app_port
+#
+# List our ports here using the forman array_name[app_name]port
+app_port[authelia]=9091
+app_port[filebrowser]=9090
+app_port[jackett]=9117
```
flying-sausages, line 8: I swear we had at least two more apps hmmmm

userdocs: I think where the array is stored matters more. `/swizzin/ports` or something. App only need ot added as the scripts are updated.

PR #629, opened 2021-04-13
feat!: bash completion

userdocs:
## Description

This pull request adds bash completion for the `box` command.

## Fixes issues: 
- na

## Proposed Changes:
- bash completion for the box command

## Change Categories
- New feature 

## Checklist

    - PR link: 
    - PR link: 

## Test scenarios

Tested on BASH Debian Buster

### Architectures

| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|	✅		|			|			|				|
| Stretch		|			|			|			|				|


scripts/update/bash_completion.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+#
+# Source the bash_completion installer
+. /etc/swizzin/sources/functions/bash_completion
```
flying-sausages, line 4: Nothing being sourced here
```suggestion
bash /etc/swizzin/sources/functions/bash_completion
```

setup.sh
```diff
@@ -375,6 +375,9 @@ function _post() {
         echo_info "You can now use the box command to manage swizzin features, e.g. \`box install nginx panel\`"
     fi
     echo_docs getting-started/box-basics
+    #
+    # Source the bash_completion installer
+    . /etc/swizzin/sources/functions/bash_completion
```
flying-sausages, line 380:
```suggestion
    # run the bash_completion installer
    bash /etc/swizzin/sources/functions/bash_completion
```

sources/functions/bash_completion
```diff
@@ -0,0 +1,18 @@
...
+    echo_info "Shell completions for the box command have been installed. They will be applied the next time you log into a bash shell"
+fi
+#
+# Create our apps and upgrade lists for the arrays.
+find "/etc/swizzin/scripts/install/" -type f -exec basename {} \; | sort | awk '{printf "%s ",$1 }' > /etc/swizzin/sources/bash_completion/completion.apps
+find "/etc/swizzin/scripts/upgrade/" -type f -exec basename {} \; | sort | awk '{printf "%s ",$1 }' > /etc/swizzin/sources/bash_completion/completion.upgrade
+#
+# Make sure the correct permissions are set
+chmod 755 '/etc/swizzin/sources/bash_completion'
+find "/etc/swizzin/sources/bash_completion/" -type f -exec chmod 644 {} \;
```
flying-sausages, line 18: Does this have to be done every single time? I think it's only necessary to do in the installer. Permissions should not change if you're `>`ing the files. The directory perms should stay the same too.

setup.sh
userdocs: does this change improve or fix anything?

sources/functions/bash_completion
userdocs: I don't see why it would be a problem to make sure the permissions are reset and correct on an update.

setup.sh
flying-sausages: https://superuser.com/questions/176783/what-is-the-difference-between-executing-a-bash-script-vs-sourcing-it

userdocs: That does not answer the question. Whether source or execute in a subshell makes not difference to this this script.

Unless there is a specific reason running this in a subshell is an improvement to just running it.

scripts/update/bash_completion.sh
```diff
@@ -0,0 +1,30 @@
...
+fi
+
+# Add all files in /var/lib/swizzin/db/bash_completion to the bash_completion_perms array to process in our for below
+readarray -t bash_completion_perms < <(find "/var/lib/swizzin/db/bash_completion" -type f -exec basename {} \;)
+# Using a for and the bash_completion_perms array check through all files in the /var/lib/swizzin/db/bash_completion have 644 perms set otherwise do nothing
+for file_names in "${bash_completion_perms[@]}"; do
+    if [[ "$(stat -c '%A' "/var/lib/swizzin/db/bash_completion/${file_names}")" != '-rw-r--r--' ]]; then
+        chmod 644 "/var/lib/swizzin/db/bash_completion/${file_names}"
+    fi
+done
```
flying-sausages, line 30: I added `swizdb list` in #633 which I think would be super useful for this, you should be able to just use that for the forloop in a oneliner

scripts/update/bash_completion.sh
```diff
@@ -0,0 +1,30 @@
...
+swizdb set bash_completion/upgrade.list "$(find "/etc/swizzin/scripts/upgrade/" -type f -exec basename {} \; | sort)"
+
+# Check to make sure the completion file /etc/swizzin/sources/bash_completion is set to 644
+if [[ $(stat -c '%A' /etc/swizzin/sources/bash_completion) != '-rw-r--r--' ]]; then
+    chmod 644 '/etc/swizzin/sources/bash_completion'
+fi
+
+# Check to make sure the directory /var/lib/swizzin/db/bash_completion is set to 755
+if [[ $(stat -c '%A' /var/lib/swizzin/db/bash_completion) != 'drwxr-xr-x' ]]; then
+    chmod 755 '/var/lib/swizzin/db/bash_completion'
```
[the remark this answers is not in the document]
flying-sausages: Might be worth checking permissions of directories above as well if we want to be thorough about it.

setup.sh
```diff
@@ -375,6 +375,12 @@ function _post() {
         echo_info "You can now use the box command to manage swizzin features, e.g. \`box install nginx panel\`"
     fi
     echo_docs getting-started/box-basics
+    #
+    # Run the bash_completion installer from the update folder
+    if ! bash /etc/swizzin/scripts/update/bash_completion.sh; then
```
flying-sausages, line 380: As it is now, this will always return true.

As far as I can see, there's nothing inside of `bash_completion.sh` in the likes of `exit 1` which would then send the escape code back for the `bash` call to return. As it is right now, if there's a failure in that update script, this condition will not trigger.

I'm also wondering what honestly _could_ fail, and I don't see that much... might as well remove the if statement here in my opinion.

liaralabs: This is just pulled over from the dependencies example which I pointed userdocs to in order to help get the PR as this snippet falls somewhere between function and snippet. You *could* set `-e` on the completion script and then if any command errors at all the script will exit. Is this the behavior we want though?

flying-sausages: aha, I see. If the bash_completions fail, I don't think the whole installer should exit out.

liaralabs: Since this is running on a bash subshell (either run the command as `bash -e ...` or `set -e` at the top of the snippet), only the running shell should exit

flying-sausages: Oh, sorry, I was referring to the exit two lines below this one (L382). That one would exit the whole `setup.sh`

userdocs: It was from this, the only other example of this behaviour.

https://github.com/swizzin/swizzin/blob/master/setup.sh#L205-L208

flying-sausages: Ah I see, the `apt` functions there are the ones that will throw an `exit 1`, which is then forwarded back to the `bash` on those lines, which will then throw that echo and kill the script.

setup.sh
```diff
@@ -375,6 +375,12 @@ function _post() {
...
+        echo_error "bash_completion update script failed."
+        exit 1
+    fi
```
flying-sausages, line 383:
```suggestion
    bash /etc/swizzin/scripts/update/bash_completion.sh
```

flying-sausages (approved):

PR #630, opened 2021-04-15
fix: remove fortune dependency

flying-sausages:
## Description
Removes `fortune` from dependencies

## Proposed Changes:
- remove `fortune` from dependencies

## Change Categories

- Bug fix 

## Checklist

    - PR link: 
    - PR link: 

### Architectures

| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|

### DIFF
--- a/scripts/update/10-dependencies.sh
+++ b/scripts/update/10-dependencies.sh
@@ -39,7 +39,7 @@ if [[ $trigger_apt_update == "true" ]]; then
 fi
 
 #space-separated list of required GLOBAL SWIZZIN dependencies (NOT application specific ones)
-dependencies="whiptail git sudo curl wget lsof fail2ban apache2-utils vnstat tcl tcl-dev build-essential dirmngr apt-transport-https bc uuid-runtime jq net-tools fortune gnupg2 cracklib-runtime"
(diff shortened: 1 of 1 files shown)


bakerboy448 (approved):
I mean... is there really anything to review here. ;)  it is no longer in dependencies 

1 question, (do we want/need | should we) remove it if it exists?

flying-sausages:
> (do we want/need | should we) remove it if it exists?

I would say no.

We'd have to do this through an `update` function. That would mean we would have to define a condition by which we'd need to determine if it is installed. This could be a bad condition, because there can be a case someone either a) wants it to stay or b) installed it manually. We could detect option b, but it would be difficult as we'd have to check `apt` logs, and those could be removed or just rolled off. We cannot really detect option a.

For this reason, we do not remove packages previous dependencies like this. We would simply announce this, tell people how they can do it themselves. This has question has been asked by me in Discord before, and this is the conclusion we came to  back then. It's been over a year so I doubt I can pull a link up.

Also, the package is ~100kB, and it is harmless due to its... nature if you can call it that.

liaralabs:
The package never actually got installed because it should have been `fortune-bin` thus worrying about removal anyway would be relatively moot.

PR #622, opened 2021-04-10
fix(sonarr)!: deprecate v2

bakerboy448:
## Description

- Removes Sonarr v2 scripts
- Updates Sonarr v2 to v3  (can discuss if needed; if so then we keep the upgrade script and do not have  a sonarr update script)

<details>
<summary>Label of collapsible goes here</summary>

</details>

## Fixes issues: 
- Fixes #596

## Proposed Changes:
- see above

## Change Categories

- Breaking change 
- Other

## Checklist

- [x] Docs have been made OR are not necessary;  - need to discuss what's happening with docs
    - PR link: https://github.com/liaralabs/docs.swizzin.ltd/pull/68
- [x] Changes to panel have been made OR are not necessary - do we want to remove the legacy code yet?
    - PR link: https://github.com/liaralabs/swizzin_dashboard/pull/34
- [X] Code is formatted [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling) - literal copy/paste of existing
- [x] Shellcheck isn't screaming [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling) - hopefully
- [X] Prints to terminal are handled [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#printing-into-the-terminal)
- [X] I have commented my code, particularly in hard-to-understand areas

## Test scenarios

if this doesn't work i'd be shocked.
### Architectures

|   			| `amd64` 	| `armhf` 	| `arm64` 	| Unspecified 	|
|--------		|-------- 	|-------- 	|-------- 	|----------		|
| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|
| Raspbian  	|	⚫️		|			|	⚫️		|	⚫️			|

PR #680, opened 2021-05-28
feat: consistent permission checking

oBusk:
## Description

Seems some bash scripts has the incorrect permissions checked in.

Whenever I run `box upgrade nginx` I got 

```bash
...     Reinstalling nginx config for sonarr
/usr/local/bin/swizzin/upgrade/nginx.sh: line 53: /usr/local/bin/swizzin/nginx/sonarr.sh: Permission denied
        ✔   Done
```

So I looked at [`nginx/sonarr.sh`](https://github.com/swizzin/swizzin/blob/4a9cd50bf9397468987272ee5e342feb2f6d557d/scripts/nginx/sonarr.sh) and it only had `-rw-r--r--` while all others had `-rwxr-xr-x`. I guess the permissions was not set correctly in the recent #622. 

Also fixed permissions on some other scripts that also had incorrect permissions.

## Fixes issues: 
- Permission denied when running `box upgrade nginx`

## Proposed Changes:
- Fixes permissions on multiple script files

## Change Categories

- Bug fix 

## Checklist

    - PR link: 
    - PR link: 
- [x] I have functionally tested the changes being introduced

### Architectures

| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|


oBusk:
I looked a bit close at the code and noticed that the permissions of the entire [scripts directory is set after cloning in setup.sh](https://github.com/swizzin/swizzin/blob/4a9cd50/setup.sh#L53), so I figured we probably want to do the same when updating from the repo to? I added this to the box update function, but not sure if that is the correct place.

oBusk:
I created a separate PR #681 for adding a simple github action to check the permissions of all scripts to help be consistent with script files in the future. Even if I added the chmod to the update script

flying-sausages:
Good catch, thanks!

So here's a little structural dilemma we have with `box update`. As git will pull in the new changes, nothing inside of the `box` file itself will be ran on the first update. This means a user will need to run `box update` to apply a change like this.

I would suggest to move the line into a new/exisitng prioritised script inside of `scripts/update/`  (maybe at the top of `checkreqs` script?) as those files are only ran through _after_ the git update. This way it will apply the next time box runs.

oBusk:
Feels strange to do chmod stuff in a script that is named for checking requirements? Do you feel ok with having it in a script that runs before `checkreqs`?

scripts/update/04-chmod.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Ensures all scrips has permissions so that they are executable
+
+echo_info "Setting permissions of scripts"
+# Set permissions of all scripts
+chmod -R 700 /etc/swizzin/scripts
+
+echo_info "Setting permissions of sources"
+# Sources are only `.` or `source`d and does not need to be executable
+chmod -R 600 /etc/swizzin/sources
```
flying-sausages, line 10: I think this will take a couple  miliseconds, an info is too prominent in this instance and we can slap them all under one "go"
```suggestion
echo_progress_start "Checking permissions on swizzin files"

chmod -R 700 /etc/swizzin/scripts # Set permissions of all scripts
chmod -R 600 /etc/swizzin/sources # Sources are only `.` or `source`d and does not need to be executable

echo_progress_done "Permissions changed"
```

flying-sausages: we might need `+rx` on directories under `sources/`, e.g. `sources/functions`.

scripts/update/04-chmod.sh
```diff
@@ -3,7 +3,8 @@
 
 echo_progress_start "Checking permissions on swizzin files"
 
-chmod -R 700 /etc/swizzin/scripts # Set permissions of all scripts
-chmod -R 600 /etc/swizzin/sources # Sources are only `.` or `source`d and does not need to be executable
+chmod -R 700 /etc/swizzin/scripts        # Set permissions of all scripts
+chmod -R 600 /etc/swizzin/sources        # Sources are only `.` or `source`d and does not need to be executable
+chmod 700 /etc/swizzin/sources/functions # Best to keep X permission on the folder inside sources
```
flying-sausages, line 8: Uhhh this should work, no? If I can make sure I can forget about this file and not keep the hardcoded dirs up to date I won't mind
```suggestion
find /etc/swizzin/sources -type d -exec chmod 700 {} \; # Best to keep X permission on the folder inside sources
```

oBusk: Aaaah of course! that's great.

oBusk:
To reduce the number of scenarios I thought it'd be good if the same mechanics was used when setting up as when updating. So I moved the chmod into a `script` and run it both when setting up and updating. Just an idea I had!

flying-sausages:
> To reduce the number of scenarios I thought it'd be good if the same mechanics was used when setting up as when updating. So I moved the chmod into a `script` and run it both when setting up and updating. Just an idea I had!

Running it on `setup.sh` makes sense but we have a policy of the format for files in `sources/functions` which basically says no file in that dir should execute on its own when sourced. You can read about that here https://swizzin.ltd/dev/structure/#functions

We already have something similar happening with the way we install dependencies in `setup.sh` and I think it's fine to execute that script directly if it's in the update folder as it was before

flying-sausages:
Sorry about the amount of changes and remarks, but I am the human-like AI CI system of this repo hahaha

oBusk:
> > To reduce the number of scenarios I thought it'd be good if the same mechanics was used when setting up as when updating. So I moved the chmod into a `script` and run it both when setting up and updating. Just an idea I had!
> 
> Running it on `setup.sh` makes sense but we have a policy of the format for files in `sources/functions` which basically says no file in that dir should execute on its own when sourced. You can read about that here https://swizzin.ltd/dev/structure/#functions
> 
> We already have something similar happening with the way we install dependencies in `setup.sh` and I think it's fine to execute that script directly if it's in the update folder as it was before

Yeah I figured that `functions` should be functions and not executable scripts, so I put it in [`scripts/setpermissions`](https://github.com/swizzin/swizzin/pull/680/files#diff-b666b5988ec997ef616a929f7c150db8e53bb4f548fe875d8b364781c4f206cb). That looked sensible. But if you would prefer with `. /etc/swizzin/update/04-chmod.sh` inside `setup.sh` I'm fine with that too.

> Sorry about the amount of changes and remarks, but I am the human-like AI CI system of this repo hahaha

😂 Don't apologize for checking the changes I'm suggesting. That makes me confident that I'm actually doing something that is wanted in the end and shit won't implode because something I don't understand happens down the line 😛

flying-sausages:
Oh! I misread, I thought it was in functions after I saw it sourced. Yeah, let's keep it in the updater dir, one less file to track. We also generally prefer to run the files through `bash` instead of sourcing for legibility and _permission issues_ hahaha

flying-sausages:
image
is this just a me thing? I can change the permissions of these files from the CLI directly just fine...

flying-sausages:
Ok I don't understand _why_ this fixes it, but might as well...

flying-sausages (approved):
I'm ok with this being merged into `develop` as long as @liaralabs is ok with the permissions of ~20 files being changed.

liaralabs (approved):
Looks good to me, unified permissions is a good thing 👍

flying-sausages:
Congrats :) feel free to come and claim your Contributor role in our Discord!

PR #681, opened 2021-05-28
feat(ci): Enforce correct script permissions

oBusk:
## Description

As part of #680 some scripts had inconsistent permissions checked in. I figured a simple shell script could be used to test this so that it's easy to see when pull requests creates files that has incorrect permissions.

## Fixes issues: 
- Makes it harder to accidentally check in script files with incorrect permissions.

## Proposed Changes:
- Add a very simple github action to check the permissions on every push and pull request

## Change Categories

- Contribution flow 

## Checklist

    - PR link: 
    - PR link: 
- [x] I have functionally tested the changes being introduced

## Test scenarios

I've tested that it fails when there are scripts that does not have `755` and that it succeeds when they ahve `755`.


flying-sausages:
ooh hell yea

Could you move the script into the CI job itself? Otherwise we could also make a dir like `.github/scripts/` to store CI helper scripts

oBusk:
@flying-sausages should we maybe check permisisos on any other files? Like `sources/functions/*`, `setup.sh`, `scripts/{box, nukeovh, php-fpm-cli, rtx, setdisk}`, and `sources/bash_completion`, from what I can see.

flying-sausages:
- `functions` are always `source`d or `. `ed into code, so they just need `o+r`
- `setup.sh` only gets ran once ever from the install command (which curls it to bash) so that could as well be `000` hahaha
- anything in `scripts/` **should indeed be checked** for `o+rwx`, including those one-offs you mention
- `bash_completions` already has its own check in its associated updater functions

oBusk:
Yeah I figured the `functions` wouldn't necessarily need `x`, but just saw they were a bit inconsistent as of now. some of them have `x` but some dont. But I guess it doesn't really matter, so checking/enforcing `o+r` is maybe just unecessary.

oBusk:
I opted for `.github/scripts` over inline bash script btw.

1. Very easy to run over and over for testing :P 
2. While this particular check is very small, I imagine there could be other, more complex scripts that oculd be put into it.

Not sure if it is necessary structure to have separate workflow-file for every job. Right now I've configured to run on `push` and `pull_request`. Feels like `sh-checker.yml` also is _linting_. Do you prefer to only run these types of checks on PRs and not push? Either way, I feel like we could merge `sh-checker.yml` and `check-permissions` into a single workflow file with 2 separate jobs. `ci.yml`, `testing.yml`, `linting.yml` or smth

flying-sausages:
`sh-checker.yml` only does formatting right now, not linting. We have _way_ too many linting errors we need to go through first, so it just ends up being a permanent failure indicator and I don't want to have that "fatigue". We had codacy and one other thing to do linting but those somehow disappeared?

We can keep this a little messy until we have more CI jobs and a clear structure will emerge :) I'd rather merge this now (given #680 gets merged to `develop` first so we can merge it back back here and re-run the job) and re-structure CI on another PR.

oBusk:
Since more files was made consistent in #680 I started checking more files and folders to keep them consistent going forward

PR #692, opened 2021-06-01
Staging Develop to Master

flying-sausages:


scripts/nginx/prowlarr.sh
```diff
@@ -0,0 +1,82 @@
...
+  <UrlBase>$app_baseurl</UrlBase>
+  <Branch>$app_branch</Branch>
+</Config>
+PROWLARR
+
+chown -R "$user":"$user" "$app_configdir"
+
+# Switch app back off if it was dead before; otherwise start it
+if [[ $isactive != "active" ]]; then
+    systemctl stop "$app_servicefile" -q
```
liaralabs, line 79: I don't understand the purpose of this code. If it was dead before, why do we need to stop it? There is no start command before this code, only underneath this else block.

scripts/tests/nextcloud.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+#shellcheck source=sources/functions/tests
+. /etc/swizzin/sources/functions/tests
+
+check_nginx "librespeed" || BAD="true"
```
liaralabs, line 6: I don't think this is the correct nginx?

liaralabs (changes requested):

scripts/nginx/prowlarr.sh
flying-sausages: This is probably a remnant of a refactor which required the API to be up, therefore the service had to be manually brought _up_ in order for the API to be available, and therefore the bit above remained as the service might have been intentionally disabled to begin with.

But yeah it can be removed, essentially ha

scripts/tests/nextcloud.sh
flying-sausages: You think correct, ha
```suggestion
check_nginx "nextcloud" || BAD="true"
```

scripts/nginx/prowlarr.sh
```diff
@@ -0,0 +1,82 @@
...
+else
+    systemctl start "$app_servicefile" -q
+fi
```
flying-sausages, line 82:
```suggestion
if [[ $isactive == "active" ]]; then
    systemctl start "$app_servicefile" -q
fi
```

liaralabs (approved):

PR #727, opened 2021-07-23
feat: JDownloader

Aethaeran:
## Fixes issues: 
- Having to install JDownloader 2 manually.
- It is a requested feature on swizzin's feathub. `https://feathub.com/liaralabs/swizzin/+79`

## Proposed Changes:
- Adding ".idea" folder to .gitignore for JetBrains IDE users.
- Adding install, remove and test scripts for JDownloader 2

## Change Categories

- New feature 
- Contribution flow 

## Checklist

    - PR link: 
    - PR link: 
- [ ] Testing was done
   - [ ] Tests executed

### Key
✅ = Works successfully
❎ = Does not work BUT is handled gracefully
🛠 = Still WIP
❌ = Broken / not working
### Scenarios
- Fresh app install with correct JD information
    - ✅ With only one user 
    - 🛠 With multiple users present 
Currently this passes the details from the first user too. This isn't intended. Needs a fix.
- ✅ Fresh app install with incorrect JD information
    - ✅ With only one user 
    - ✅ With multiple users present
- Fresh app install with correct details for one user, and incorrect details for another user.
- 🛠 Bypass MyJDownloader detail passing entirely with environment variable
- Remove Install
    - With only one user
    - With multiple users present
- `box`  adduser\deluser adds\removes user installation. 
Simply handled by adding JDownloader to the lists of other apps already handled in a similar fashion in `box` itself.
    - `box adduser`
    - `box deluser`
- ❎ Update\Upgrade from version in master 
Deemed unnecessary due to JDownloader handling updates internally.
- ❎ Password gets changed from `box` in app 
This isn't possible. JDownloader2's Web-UI is intended to be used through their my.jdownloader.org service. The odds that the swizzin user data is going to align with your account details on their site is unlikely. The username for my.jdownloader.org is always in the form of an e-mail. Furthermore, JDownloader saves your my.jdownloader.org details in plain text in the cfg directory. We wouldn't want to have unhashed swizzin passwords just lying around.
image
### Architectures Tests

| Focal 		|	🛠		|	🛠		|	🛠		|	🛠			|
| Bionic		|	🛠		|	🛠		|	🛠		|	🛠			|
| Buster		|	🛠		|	🛠		|	🛠		|	🛠			|
| Stretch		|	🛠		|	🛠		|	🛠		|	🛠			|
| Raspbian  	|	🛠		|	🛠		|	🛠		|	🛠			|

### DIFF
--- a/scripts/install/jdownloader.sh
+++ b/scripts/install/jdownloader.sh
@@ -5,6 +5,7 @@
 ##########################################################################
 # References
 ##########################################################################
+
 # Liara already made a doc for installing JDownloader manually:
 # https://docs.swizzin.net/guides/jdownloader/
 # swizzin docs
@@ -25,8 +26,10 @@
 ##########################################################################
 # Import Sources
 ##########################################################################
-. /etc/swizzin/sources/functions/utils
+
+. /etc/swizzin/sources/functions/java
 . /etc/swizzin/sources/functions/jdownloader
+. /etc/swizzin/sources/functions/utils
 
 ##########################################################################
 # Functions
@@ -38,8 +41,8 @@ function install_jdownloader() {
     JD_HOME="/home/$user/jd2"
     mkdir -p "$JD_HOME"
 
-    # An environment variable 'MYJD_BYPASS' to bypass the following block. For unattended installs.
-    if [[ -n $MYJD_BYPASS ]]; then
+    # TODO: Test environment variable 'MYJD_BYPASS' to bypass the following block. For unattended installs.
+    if [[ ! -z $MYJD_BYPASS ]]; then
         if ask "Do you want to inject MyJDownloader details for $user?" N; then
             inject="true"
             echo_info "Injecting MyJDownloader details for $user"
@@ -72,12 +75,13 @@ function install_jdownloader() {
         cp "/tmp/JDownloader.jar" "$JD_HOME/JDownloader.jar"
     fi
 
-    command="java -jar $JD_HOME/JDownloader.jar -norestart"
-
     # TODO: Currently, we need something here to disable all currently running JDownloader installations, or the MyJD verification logic will cause a loop. Would rather we didn't.
     for each_user in "${users[@]}"; do # disable all instances
         systemctl disable --now "jdownloader@$each_user" --quiet
     done
+
+    # TODO: Figure out if this would be a WHOLE lot simpler if I didn't run the `-norestart` flag *facepalm*
+    command="java -jar $JD_HOME/JDownloader.jar -norestart"
 
     echo_progress_start "Attempting JDownloader2 initialisation"
     end_initialisation_loop="false"
@@ -98,10 +102,12 @@ function install_jdownloader() {
         trap "kill $pid 2> /dev/null" EXIT # Set trap to kill background process if this script ends.
         process_died="false"
         while [[ $process_died == "false" ]]; do # While background command is still running...
+
             echo_info "Background command is still running..." # TODO: This should be removed at PR end.
             sleep 1 # Pace this out a bit, no need to check what JDownloader is doing more frequently than this.
             # If any of specified strings are found in the log, kill the last called background command.
             if [[ -e "$tmp_log" ]]; then
+
                 if grep -q "Create ExitThread" -F "$tmp_log"; then # JDownloader exited gracefully on it's own. Usually this will only happen first run.
                     echo_info "JDownloader exited gracefully." # TODO: This should be echo_log_only at PR end.
                 fi
@@ -131,6 +137,7 @@ function install_jdownloader() {
                     fi
 
                     # This only works for verification if it is the first JDownloader instance to attempt connecting to MyJDownloader. I assume other instances use the same HTTP server.
+
                     if grep -q "Start HTTP Server" -F "$tmp_log"; then
                         echo_info "MyJDownloader account details verified."
                         kill_process="true"
@@ -163,7 +170,8 @@ function install_jdownloader() {
 }
 
 _systemd() {
-    # JDownloader will automatically create a pidfile when running. That way, systemd can use it to ensure it is disabling the correct process.
+    # JD_HOME should be setting the working directory for this instance of JDownloader.
+    # JDownloader will automatically create a pid file when running. That way, systemd can use it to ensure it is disabling the correct process.
     cat >/etc/systemd/system/jdownloader@.service <<EOF
 [Unit]
 Description=JDownloader Service
@@ -185,28 +193,27 @@ EOF
 # Script Main
 ##########################################################################
 
-if [[ -n "$1" ]]; then # Install jd2 for user that was passed to script as arg (i.e. box adduser <user>) and do not execute the rest
+if [[ -n "$1" ]]; then # Install JDownloader for JUST the user that was passed to script as arg (i.e. box adduser $user)
     user="$1"
     install_jdownloader
     exit 0
 fi
 
-#shellcheck source=sources/functions/java
-. /etc/swizzin/sources/functions/java
-install_java8
+install_java8 # Install Java as it is a dependency.
 
-_systemd
-
-readarray -t users < <(_get_user_list)
-for user in "${users[@]}"; do # Install a separate instance for each user
+readarray -t users < <(_get_user_list) # Install a separate JDownloader instance for each user
+for user in "${users[@]}"; do
     install_jdownloader
 done
-# Don't start services until after each user is installed.
-for user in "${users[@]}"; do # Install a separate instance for each user
+
+_systemd # Insert service file to /etc/systemd/system
+
+# Don't start services until after each user is installed. Due to current verification process.
+for user in "${users[@]}"; do # Enable a separate service for each swizzin user
     echo_progress_start "Enabling service jdownloader@$user"
     systemctl enable -q --now jdownloader@"$user" --quiet
     echo_progress_done
 done
 
-touch /install/.jdownloader.lock
-echo_success "JDownloader installed"
+touch /install/.jdownloader.lock # Create lock file so that swizzin knows JDownloader is installed.
+echo_success "JDownloader installed" # Winner winner. Chicken dinner.
--- a/scripts/install/jdownloader.sh
+++ b/scripts/install/jdownloader.sh
@@ -28,7 +28,10 @@
 ##########################################################################
 
 # TODO: Make $get_most_recent_dir_in_folder function
-
+function get_most_recent_dir() {
+    #shellcheck disable=2012
+    ls -td -- "$1"/* | head -n 1
+}
 # TODO: Move this function to another file so the end user could use it to inject their details as well.
 
 function get_myjd_info() {
@@ -126,7 +129,7 @@ function install_jdownloader() {
             sleep 2 # Pace this out a bit
             # If any of specified strings are found in the log, kill the last called background command.
             if [[ -e "$tmp_log" ]]; then # If the
-                # TODO: Seems to be missing this detection if the background command closes too quickly.
+                # TODO: Seems to be missing this detection if the background command closes too quickly. Increased previous sleep to 2 seconds to see if it helps.
                 if grep -q "Create ExitThread" -F "$tmp_log"; then # JDownloader exited gracefully on it's own. Usually this will only happen first run.
                     echo_info "JDownloader exited gracefully." # TODO: This should be echo_log_only at PR end.
                     trap - EXIT   # Disable the trap on a normal exit.
@@ -137,8 +140,8 @@ function install_jdownloader() {
                     # Pretty sure this will remove the long pause.
                     keep_sleeping="true"
                     while [[ ! $keep_sleeping == "false" ]]; do
-                        if grep -q "No Console Available" -F "$tmp_log" || grep -q "Start HTTP Server" -F "$tmp_log"; then
-                            keep_sleeping="true"
+                        if [[ grep -q "No Console Available" -F "$tmp_log" || grep -q "Start HTTP Server" -F "$tmp_log" ]]; then
+                            keep_sleeping="false"
                         else
                             sleep 1 # Wait until JDownloader has attempted launching the HTTP server.
                         fi
--- a/scripts/install/jdownloader.sh
+++ b/scripts/install/jdownloader.sh
@@ -50,12 +50,20 @@ function install_jdownloader() {
 
     # TODO: Java will detect if a .jar is corrupt, we could use that to our advantage. "Error: Invalid or corrupt jarfile"
     echo_progress_start "Downloading JDownloader.jar"
-    if [[ ! -e "/tmp/JDownloader.jar" ]]; then
-        wget -q http://installer.jdownloader.org/JDownloader.jar -O "/tmp/JDownloader.jar" || {
+    while [[ ! -e "/tmp/JDownloader.jar" ]]; do
+        if wget -q http://installer.jdownloader.org/JDownloader.jar -O "/tmp/JDownloader.jar"; then
+            echo_info "Jar downloaded..."
+            if ! java -jar "/tmp/JDownloader.jar" 2>/dev/null;then
+                echo ".jar is corrupt. Removing, and trying again."
+                rm "/tmp/JDownloader.jar"
+            else
+                echo ".jar is valid."
+            fi
+        else
             echo_error "Failed to download"
-            exit 1
-        }
-    fi
+        fi
+    done
+
     echo_progress_done "Jar downloaded"
 
     if [[ ! -e "$JD_HOME/JDownloader.jar" ]]; then
--- a/scripts/install/jdownloader.sh
+++ b/scripts/install/jdownloader.sh
@@ -62,6 +62,7 @@ function install_jdownloader() {
 
     if ask "Do you want to inject MyJDownloader details for $user?" N; then
         inject="true"
+        echo_info "Injecting MyJDownloader details for $user"
         get_myjd_info # Get account info for this user. and insert it into this installation
     else
         inject="false"
@@ -102,14 +103,14 @@ function install_jdownloader() {
                     # TODO: Pretty sure I can do this without the long pause here... Somehow
                     sleep 15 # Wait this long to make sure JDownloader has gotten to the point of attempting to launch HTTP server.
                     if grep -q "No Console Available" -F "$tmp_log"; then
-                        echo_warn "MyJDownloader account details were incorrect. Please try again."
+                        echo_warn "MyJDownloader account details were incorrect."
                         if [[ $inject == "true" ]]; then
-                            echo
+                            echo_info "Please enter the MyJDownloader details again."
                             get_myjd_info # Get account info for this user. and insert it into this installation
                         else
                             end_loop="true"
-                        kill_process="true"
                         fi
+                        kill_process="true"
                     fi
                     # TODO: This said verified regardless that new >.< Need a diffrenet string to verify with. "FINER [ org.appwork.loggingv3.LogV3(finer) ] -> Load Translation file:"
                     # TODO: This only works for verification if it is the first JDownloader instance to attempt connecting to MyJDownloader. I assume other instances use the same HTTP server.
--- a/scripts/install/jdownloader.sh
+++ b/scripts/install/jdownloader.sh
@@ -25,16 +25,14 @@
 ##########################################################################
 # Import Sources
 ##########################################################################
-. /sources/functions/utils
-. /sources/functions/jdownloader
+. etc/swizzin/sources/functions/utils
+. etc/swizzin/sources/functions/jdownloader
 
 ##########################################################################
 # Functions
 ##########################################################################
 
 function install_jdownloader() {
-    
-    . /sources/functions/utils
 
     echo_info "Setting up JDownloader for $user"
     JD_HOME="/home/$user/jd2"
--- a/scripts/install/jdownloader.sh
+++ b/scripts/install/jdownloader.sh
@@ -33,6 +33,8 @@
 ##########################################################################
 
 function install_jdownloader() {
+    
+    . /sources/functions/utils
 
     echo_info "Setting up JDownloader for $user"
     JD_HOME="/home/$user/jd2"
--- a/scripts/install/jdownloader.sh
+++ b/scripts/install/jdownloader.sh
@@ -63,7 +63,7 @@ function install_jdownloader() {
 
     #TODO: Add environment variable to bypass the following block. For unattended installs.
 
-    if [[ NO_MYJD == "false" ]]; then
+    if [[ $NO_MYJD == "false" ]]; then
         if ask "Do you want to inject MyJDownloader details for $user?" N; then
             inject="true"
             echo_info "Injecting MyJDownloader details for $user"
@@ -71,6 +71,7 @@ function install_jdownloader() {
         else
             inject="false"
         fi
+    fi
 
     # TODO: Have this store the first downloader JDownlaoder in /opt/jdownloader, all further instances can just copy it from there.
     # TODO: JDownloader will detect if this is corrupt, we could use that to
--- a/scripts/install/jdownloader.sh
+++ b/scripts/install/jdownloader.sh
@@ -22,21 +22,19 @@
 # https://superuser.com/questions/402979/kill-program-after-it-outputs-a-given-line-from-a-shell-script
 # https://board.jdownloader.org/showthread.php?t=81433
 
-
 ##########################################################################
 # Functions
 ##########################################################################
 
-# TODO: Make $get_most_recent_dir_in_folder function
 function get_most_recent_dir() {
     #shellcheck disable=2012
-    ls -td -- "$1"/* | head -n 1
+    find $1 -mindepth 1 -maxdepth 1 -type d
 }
-# TODO: Move this function to another file so the end user could use it to inject their details as well.
+# TODO: Move this function to sources/functions/jdownloader so the end user could use it to inject their details as well.
 
-function get_myjd_info() {
+function inject_myjdownloader() {
 
-    # TODO: The way this is currently handled causes information to be passed to EACH user that it is installed for.
+    # TODO: Make this only pass env file details to master user
 
     echo_info "An account from https://my.jdownloader.org/ is required in order to access the web UI.\nUse a randomly generated password at registration as the password is stored in plain text"
     if [[ -z "${MYJD_EMAIL}" ]]; then
@@ -78,21 +76,20 @@ EOF
 function install_jdownloader() {
 
     echo_info "Setting up JDownloader for $user"
-    # TODO: Should we move this to .config instead?
-    $JD_HOME="/home/$user/jd2"
+    JD_HOME="/home/$user/jd2"
 
     if [[ $BYPASS_MYJDOWNLOADER == "false" ]]; then
         if ask "Do you want to inject MyJDownloader details for $user?" N; then
             inject="true"
             echo_info "Injecting MyJDownloader details for $user"
-            get_myjd_info # Get account info for this user. and insert it into this installation
+            inject_myjdownloader # Get account info for this user. and insert it into this installation
         else
             inject="false"
         fi
     fi
 
-    # TODO: Have this store the first downloader JDownlaoder in /opt/jdownloader, all further instances can just copy it from there.
-    # TODO: JDownloader will detect if JDownlaoder.jar is corrupt, we could use that to our advantage.
+    # TODO: Have this store the first JDownloader.jar in /tmp/, all further instances can just copy it from there.
+    # TODO: JDownloader will detect if JDownloader.jar is corrupt, we could use that to our advantage.
     echo_progress_start "Downloading JDownloader.jar"
     mkdir -p "$JD_HOME"
     if [[ ! -e "$JD_HOME/JDownloader.jar" ]]; then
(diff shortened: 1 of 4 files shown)


scripts/install/jdownloader.sh
```diff
@@ -0,0 +1,117 @@
...
+# https://swizzin.ltd/dev/structure/
+# https://www.digitalocean.com/community/tutorials/how-to-install-java-with-apt-on-debian-10
+# https://support.jdownloader.org/Knowledgebase/Article/View/install-jdownloader-on-nas-and-embedded-devices
+# Liara already made a doc for installing JDownloader manually https://docs.swizzin.net/guides/jdownloader/
+# https://linuxize.com/post/how-to-check-if-string-contains-substring-in-bash/
+# https://linuxize.com/post/bash-check-if-file-exists/
+
+# TODO: Figure out if these imports are even necessary
+. /etc/swizzin/sources/globals.sh
+. /etc/swizzin/sources/functions/users
```
flying-sausages, line 15: Not necessary, these get pulled in whenever `box` is ran and are exported into the shell for the whole `box` run

scripts/install/jdownloader.sh
```diff
@@ -0,0 +1,117 @@
...
+
+# Functions
+function install_jdownloader() {
+
+    # Get my.jdownloader info for user
+    # TODO: Should double check to confirm everything is accurate, and loop back if anything isn't filled out.
+    # TODO: swizzin likely has utils for this already
+    echo_query "The user will require an account at https://my.jdownloader.org/ in order to access your JDownloader installation's web UI.\nIt is recommended to use a randomly generated password for the account since the password is save in plain text in the program folder.\nEnter the e-mail used to access this account once one is created:"
```
flying-sausages, line 23: I'd break this up a bit
```suggestion
    echo_info "The user will require an account at https://my.jdownloader.org/ in order to access your JDownloader installation's web UI.\nIt is recommended to use a randomly generated password for the account since the password is save in plain text in the program folder."
    echo_query "Enter the e-mail used to access this account once one is created:"
```

scripts/install/jdownloader.sh
```diff
@@ -0,0 +1,117 @@
...
+    read -r 'myjd_email'
+    echo_query "Please enter the password for the account."
+    read -r 'myjd_password'
+    echo_query "Please enter the device name they would like this JDownloader installation to show up as."
```
flying-sausages, line 27:
```suggestion
    echo_query "Please enter the desired device name"
```

scripts/install/jdownloader.sh
```diff
@@ -0,0 +1,117 @@
...
+    read -r 'myjd_devicename'
+
+    # Install JDownloader
+    echo_progress_start "Downloading and installing JDownloader for $user"
+    mkdir -p /home/"$user"/jd
+    wget -q http://installer.jdownloader.org/JDownloader.jar -O /home/"$user"/jd/JDownloader.jar
```
flying-sausages, line 33: Do we need a whole `jar` for each user? why not use one `jar` and have all the services refer to that bin?

scripts/install/jdownloader.sh
```diff
@@ -0,0 +1,117 @@
...
+    SUB='not found'
+    if [[ "$STR" == *"$SUB"* ]]; then
+        echo_info "Java was not installed correctly. Check the swizzin log for details. Exiting."
+        exit 1
+    else
+        echo_info "Java was installed successfully."
+    fi
+else
+    echo_info "Java was found. No need to install."
+fi
```
flying-sausages, line 88: use the `java` functions we already have in `sources/functions/java`, I think they're merged in with airsonic

scripts/install/jdownloader.sh
```diff
@@ -0,0 +1,117 @@
...
+Environment=JD_HOME=/home/%i/jd
+ExecStart=/usr/bin/java -Djava.awt.headless=true -jar /home/%i/jd/JDownloader.jar
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+# Check for all current swizzin users, and install JDownloader for each user.
+users=("$(_get_user_list)")
+
```
flying-sausages, line 109: Missing a closing `echo_progress_done` here to match the start from L91

flying-sausages: The for loop call on L111 will make all sorts of prints, `starts` need to be directly closed by `done`

scripts/tests/jdownloader.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+# TODO
```
flying-sausages, line 2: If you delete this file and `box test jdownloader` runs just fine after an install, you don't even need this. The default tests can go far, test files are only needed for exceptions to the defaults.

flying-sausages (commented):
Quick read through while I'm on vacay, I'll give this a shot sometime at the end of the week maybe

scripts/install/jdownloader.sh
Aethaeran: Yeah I get way too wordy. Acceptable change in my books

scripts/install/jdownloader.sh
Aethaeran: Also a good call. Honestly still haven't looked at the differences between the different swizzin echos. But this another time where I get a bit wordy. I'm sure this could be condensed further still.

scripts/install/jdownloader.sh
Aethaeran: I had a feeling if I left a TODO about that someone would chime in for me. Appreciated!

scripts/install/jdownloader.sh
```diff
@@ -20,11 +20,12 @@ function install_jdownloader() {
     # Get my.jdownloader info for user
     # TODO: Should double check to confirm everything is accurate, and loop back if anything isn't filled out.
     # TODO: swizzin likely has utils for this already
-    echo_query "The user will require an account at https://my.jdownloader.org/ in order to access your JDownloader installation's web UI.\nIt is recommended to use a randomly generated password for the account since the password is save in plain text in the program folder.\nEnter the e-mail used to access this account once one is created:"
+    echo_info "The user will require an account at https://my.jdownloader.org/ in order to access your JDownloader installation's web UI.\nIt is recommended to use a randomly generated password for the account since the password is save in plain text in the program folder."
```
flying-sausages, line 23:
```suggestion
    echo_info "An account from https://my.jdownloader.org/ is required in order to access the web UI.\nUse a randomly generated password at registration as the password is stored in plain text."
```

Aethaeran: Thank you!

scripts/install/jdownloader.sh
Aethaeran: b said something similar on the Discord. I will certainly be looking into this once everything else is said and done. I'm not 100% sure which files can be duplicated across installations. Although we would certainly need the `cfg` directory generated by each intallation to be separate. As there is an absolute plethora of options for JDownloader, and at the very minimum each users account information for accessing the web ui via the third party would have to be self contained.

It's quite possible that I could do something like drop the JDownloader.jar into /opt/jdownloader instead of to each user's home directory. The service file passes the working directory for JDownloader, so where the actual .jar is containted could be inconsequential.

However, I'm also not completely certain that that JDownloader.jar file will not be replaced at some point by an internal update. My concern here being that if one user's configuration was not updated along with the application files (for instance if they had jdownloader disabled for some time) then we might end up with some broken installations.

TL;DR - I'm not sure that having multiple copies of a 3.8 MB .jar file is something we should be worried about if it might cause more headaches down the line.

scripts/install/jdownloader.sh
Aethaeran: We did discuss this in passing on the Discord. I believe we decided to just have me do my own thing? As the java scripts for airsonic were Java 8 specific for some reason or another. I don't think JDownloader will care either way, because as far as I know it only requires `Java 1.8`.

I'm open to discuss it further if you would like.

scripts/tests/jdownloader.sh
Aethaeran: I did not realize that `box test jdownloader` would do anything without this file. I will have to look into the default operations of `box test`.

However, I am thinking I would like to add a custom test. Specifically, I'm thinking something to detect if their installation is accessible via the third party website. Since the login information for that will be stored locally in plain text anyways. This would give sysadmins a quick test to see if any of the login information was entered incorrectly on install, if their login data has changed during install, or if the installation is unaccessible.

scripts/install/jdownloader.sh
Aethaeran: I think I resolved this since my changes today. You'll have to take another look and let me know.

scripts/install/jdownloader.sh
flying-sausages: Ait, let's keep this thread open until the end of the PR and let's revisit it then

scripts/install/jdownloader.sh
flying-sausages: I don't see a reason to deviate from JRE8 and using the wheel we already invented. Maintaining an extra install of JRE as we'll have to start maintaining those in all of the service files for Java apps.

Is there any outstanding advantage when using +8, which would outweigh the added complexity to the system?

scripts/tests/jdownloader.sh
flying-sausages: Fair point that I should write docs for the testing functionality.

I would say that we should not rely on using tools from systems outside of our control, if that portal goes down we need a rewrite on our end. As long as the services are up, and they're accessible internally and via nginx, we're good I would say.

flying-sausages:
image
FYI

scripts/install/jdownloader.sh
flying-sausages: I think it would actually be nice to move some of the "testing" logic into the function defined here so that we use it anytime we install JDK

https://github.com/swizzin/swizzin/blob/master/sources/functions/java

this means we should handle the entire java install above hrough something like
```sh
. /etc/swizzin/sources/functions/java
install_jdk8
```

scripts/tests/jdownloader.sh
Aethaeran: I would rather access it directly as well. I'm not actually sure there is an alternative though. As far as I know JDownloader's web UI can **only** be used via their site. I thought I might have seen something contrary to this in passing the other day though... I'll have to take another look around.

flying-sausages: We can just check if the ports are open and being listened to, not necessarily make some `curl`/HTTP requests to them. We have a basetest function for this. I would say that's _good enough_

Aethaeran: Oh alright then. That does sound like a better solution to me. Where might I find this function?

scripts/install/jdownloader.sh
Aethaeran: This change seems fine to me. Like I said, I don't think JDownloader is going to are either way.

Except, I should note that I've moved away from:
```
STR=$(java --version)
SUB='not found'
if [[ "$STR" == *"$SUB"* ]]; then
```
because it was not functioning properly under certain conditions. I'm not sure I could replicate them but it was because `java --version` was outputting something other than "command not found" when Java was not installed, and therefore skipping installation. I think possibly the system had just recently had Java removed, but I could be mistaken.
I have been using this instead since. Haven't had any issues since. 
```
if [[ ! -e /usr/bin/java ]]; then
```
I'm not sure where you would want to put them into the `install_java8` function itself. I'm a little confused because it looks like this function only handles `buster` installs. I'm I reading this incorrectly?

scripts/tests/jdownloader.sh
flying-sausages: Check this https://github.com/swizzin/swizzin/blob/master/sources/functions/tests, let me know if the inline docs aren't good enough

scripts/install/jdownloader.sh
flying-sausages: > I'm not sure where you would want to put this logic into the install_java8 function itself. I'm a little confused because it looks like this function only handles buster installs. Am I reading this incorrectly?

It handles buster explicitly, and all others (the `*)` case) are handled through the apt install.

We _might_ want to check for **pre-existing** JRE installs before the actual install here and try to exit gracefully or offer some mitigations.

scripts/tests/jdownloader.sh
flying-sausages: ++ https://swizzin.ltd/dev/structure/#test

scripts/install/jdownloader.sh
Aethaeran: Ahhhhh see. I didn't know about `*)` case. This makes more sense now. If I remember correctly. You mentioned something about airsonic requiring Java 8 specifically for some reason or another. So for the install_java8 function, I'd recommend checking for `/usr/bin/java` first, and then check to see if a version other than 8 is installed. Then have some sort of user intervention at that point. They could choose to (A) Cancel. (B) Remove their current installation for a Java 8 install instead. (C) If this is somehow possible, opt into having parallel java installations. (All apps that _require_ `Java 8` would have to act accordingly)

Aethaeran: This is very rough, but essentially something like this.
```
if [[ ! -e /usr/bin/java ]]; then
    echo_info "Java is not installed. We are free to install Java 8"
else
    echo_info "Java is installed. Check which version"
    if [[ ! "$(/usr/bin/java --version)" == *'build 8.'* ]]; then
        echo_error "Current Java version is not 8."
        echo_info "We can handle this in three ways:"
        echo_info "(A) "
        echo_info "(B) "
        echo_info "(C) "
        echo_query "How do you want to handle this?"
        read -r "answer"
        case $answer in
            "A")
                echo_info "Do A"
                ;;
            "B")
                echo_info "Do B"
                ;;
            "C")
                echo_info "Do C"
                ;;
            *)
                echo_info "Loop back and get a proper answer."
                ;;
        esac
    fi
fi
```

Aethaeran:
> 
> 
> image
> FYI

I've been pushing my scripts from a Windows installation. They all come up 755 when I check them with git bash. I'm assuming it's being lost on push since Windows doesn't handle perms this way.

scripts/install/jdownloader.sh
flying-sausages: Hm.

Fuck it, that's for another PR. We should re-evaluate if it's mybe time to upgrade the JRE too, but that's out of scope for this app introduction. If we use the function we have for it now, then making the changes there later will apply to the whole toolset, and that will be slick.

Aethaeran: Sounds good to me. Will change it as soon as I have all the install issues sorted out

scripts/tests/jdownloader.sh
Aethaeran: Thinking the default tests aren't going to cut it in this case.
```
# box test jdownloader
INFO	Running default test for jdownloader.
	This test is likely to fail in case the item is not a 
	standard service+nginx app
...	Checking jdownloader service is active
Unit jdownloader.service could not be found.
WARN	jdownloader.service not active!
...	Checking if port jdownloader is open and listened to
WARN	Couldn't guess port
...	Checking if port jdownloader is reachable via curl
WARN	Couldn't guess port
WARN	nginx not installed, cannot verify if jdownloader is 
	reachable through it. Skipping.
ERROR	Errors were encountered
	Please consult the above and/or check the log (less -R +G 
	/root/logs/tests-2021.07.30-08:25:13.txt)
```

Aethaeran: BUT... 
I did find these today, and they suggest that it might be possible to access JDownloader without using the third party website.
[A help topic suggesting that NGINX might be an option.](https://board.jdownloader.org/showthread.php?t=76156)
[API Docs for reference](https://my.jdownloader.org/developers/)

scripts/install/jdownloader.sh
Aethaeran: Upon further review, it would seem that JDownloader really doesn't care where JD_HOME is set to (during the intial runs at least). It seems to generate the files around itself, wherever it may be being run from.

Don't have high hopes of this being an option at this point.

scripts/install/jdownloader.sh
Aethaeran: The script now installs Java with the swizzin function.
What should we call this other PR for Java changes? Should I start it? or should you?

scripts/install/jdownloader.sh
Aethaeran: I've added an awful lot of echos since this comment. Might need an additional review to see if they are excessive.

scripts/install/jdownloader.sh
Aethaeran: Just thought I'd give it a brief Google search.
[Doesn't sound promising.](https://board.jdownloader.org/showthread.php?t=74492)

scripts/tests/jdownloader.sh
flying-sausages: Yeah checking the web UI might be more trouble than it's worth honestly.

scripts/install/jdownloader.sh
flying-sausages: I'd say `fix: java management` is good enough. Feel free to have a start at it ;)

scripts/install/jdownloader.sh
flying-sausages: Fair, that pretty much says it's not possible and supported so I'm closing this one.

scripts/install/jdownloader.sh
flying-sausages: I cleaned up a majority of the echos

flying-sausages:
This is what the echos look like right now. Looks like we still need to fix a couple bugs about the intialisation sequence.

image

This is what the service dies on when started for my user
```
Jul 31 17:53:03 m1 java[88140]: 23|Log.L.log 7/31/21 3:53:03 PM - SEVERE [ Log.L.log ] -> Exception thrown at jd.SecondLevelLaunch$9$1.run(SecondLevelLaunch.java:919):
Jul 31 17:53:03 m1 java[88140]: java.lang.ExceptionInInitializerError
Jul 31 17:53:03 m1 java[88140]:         at jd.SecondLevelLaunch$9$1.run(SecondLevelLaunch.java:747)
Jul 31 17:53:03 m1 java[88140]: Caused by: java.lang.RuntimeException: No Console Available!
Jul 31 17:53:03 m1 java[88140]:         at org.appwork.console.ConsoleDialog.<init>(ConsoleDialog.java:54)
Jul 31 17:53:03 m1 java[88140]:         at org.appwork.console.ConsoleDialog.<init>(ConsoleDialog.java:60)
Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.askLoginsOnConsole(MyJDownloaderController.java:273)
Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.start(MyJDownloaderController.java:125)
Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.<init>(MyJDownloaderController.java:112)
Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.<clinit>(MyJDownloaderController.java:37)
Jul 31 17:53:03 m1 java[88140]:         ... 1 more
```

scripts/tests/jdownloader.sh
Aethaeran: You might not be wrong there. It's also entirely possible that the JDownloader logs would be telling us if it failed anyways.

Aethaeran:
> 
> 
> This is what the echos look like right now. Looks like we still need to fix a couple bugs about the intialisation sequence.
> 
> image
> 
> This is what the service dies on when started for my user
> 
> ```
> Jul 31 17:53:03 m1 java[88140]: 23|Log.L.log 7/31/21 3:53:03 PM - SEVERE [ Log.L.log ] -> Exception thrown at jd.SecondLevelLaunch$9$1.run(SecondLevelLaunch.java:919):
> Jul 31 17:53:03 m1 java[88140]: java.lang.ExceptionInInitializerError
> Jul 31 17:53:03 m1 java[88140]:         at jd.SecondLevelLaunch$9$1.run(SecondLevelLaunch.java:747)
> Jul 31 17:53:03 m1 java[88140]: Caused by: java.lang.RuntimeException: No Console Available!
> Jul 31 17:53:03 m1 java[88140]:         at org.appwork.console.ConsoleDialog.<init>(ConsoleDialog.java:54)
> Jul 31 17:53:03 m1 java[88140]:         at org.appwork.console.ConsoleDialog.<init>(ConsoleDialog.java:60)
> Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.askLoginsOnConsole(MyJDownloaderController.java:273)
> Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.start(MyJDownloaderController.java:125)
> Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.<init>(MyJDownloaderController.java:112)
> Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.<clinit>(MyJDownloaderController.java:37)
> Jul 31 17:53:03 m1 java[88140]:         ... 1 more
> ```
Just did some clean-up to the logic that should clear up the grep spam.

image

Are you not passing authentic details to the MY JD prompts? I believe that is the only reason you can get the the "No Console Available" error. That's because the headless JDownloader will try to prompt the user for the MYJD details. Since it is being run in the background, instead of a regular terminal, there is no console to request this information in, so it throws this error. I'm hoping ultimately we can use this to verify MYJD details.

Does it still say it's active like this?

image

Because when I put in accurate details I stall out somewhere around here

image

brettpetch:
> > This is what the echos look like right now. Looks like we still need to fix a couple bugs about the intialisation sequence.
> > image
> > This is what the service dies on when started for my user
> > ```
> > Jul 31 17:53:03 m1 java[88140]: 23|Log.L.log 7/31/21 3:53:03 PM - SEVERE [ Log.L.log ] -> Exception thrown at jd.SecondLevelLaunch$9$1.run(SecondLevelLaunch.java:919):
> > Jul 31 17:53:03 m1 java[88140]: java.lang.ExceptionInInitializerError
> > Jul 31 17:53:03 m1 java[88140]:         at jd.SecondLevelLaunch$9$1.run(SecondLevelLaunch.java:747)
> > Jul 31 17:53:03 m1 java[88140]: Caused by: java.lang.RuntimeException: No Console Available!
> > Jul 31 17:53:03 m1 java[88140]:         at org.appwork.console.ConsoleDialog.<init>(ConsoleDialog.java:54)
> > Jul 31 17:53:03 m1 java[88140]:         at org.appwork.console.ConsoleDialog.<init>(ConsoleDialog.java:60)
> > Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.askLoginsOnConsole(MyJDownloaderController.java:273)
> > Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.start(MyJDownloaderController.java:125)
> > Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.<init>(MyJDownloaderController.java:112)
> > Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.<clinit>(MyJDownloaderController.java:37)
> > Jul 31 17:53:03 m1 java[88140]:         ... 1 more
> > ```
> 
> Just did some clean-up to the logic that should clear up the grep spam.
> 
> image
> 
> Are you not passing authentic details to the MY JD prompts? I believe that is the only reason you can get the the "No Console Available" error. That's because the headless JDownloader will try to prompt the user for the MYJD details. Since it is being run in the background, instead of a regular terminal, there is no console to request this information in, so it throws this error. I'm hoping ultimately we can use this to verify MYJD details.
> 
> Does it still say it's active like this?
> 
> image
> 
> Because when I put in accurate details I stall out somewhere around here
> 
> image

I’d suggest use the default masteruser username and password for this. You should be able to find the _get_user_password function pretty easily.

Aethaeran:
> > > This is what the echos look like right now. Looks like we still need to fix a couple bugs about the intialisation sequence.
> > > image
> > > This is what the service dies on when started for my user
> > > ```
> > > Jul 31 17:53:03 m1 java[88140]: 23|Log.L.log 7/31/21 3:53:03 PM - SEVERE [ Log.L.log ] -> Exception thrown at jd.SecondLevelLaunch$9$1.run(SecondLevelLaunch.java:919):
> > > Jul 31 17:53:03 m1 java[88140]: java.lang.ExceptionInInitializerError
> > > Jul 31 17:53:03 m1 java[88140]:         at jd.SecondLevelLaunch$9$1.run(SecondLevelLaunch.java:747)
> > > Jul 31 17:53:03 m1 java[88140]: Caused by: java.lang.RuntimeException: No Console Available!
> > > Jul 31 17:53:03 m1 java[88140]:         at org.appwork.console.ConsoleDialog.<init>(ConsoleDialog.java:54)
> > > Jul 31 17:53:03 m1 java[88140]:         at org.appwork.console.ConsoleDialog.<init>(ConsoleDialog.java:60)
> > > Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.askLoginsOnConsole(MyJDownloaderController.java:273)
> > > Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.start(MyJDownloaderController.java:125)
> > > Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.<init>(MyJDownloaderController.java:112)
> > > Jul 31 17:53:03 m1 java[88140]:         at org.jdownloader.api.myjdownloader.MyJDownloaderController.<clinit>(MyJDownloaderController.java:37)
> > > Jul 31 17:53:03 m1 java[88140]:         ... 1 more
> > > ```
> > 
> > Just did some clean-up to the logic that should clear up the grep spam.
> > 
> > image
> > 
> > Are you not passing authentic details to the MY JD prompts? I believe that is the only reason you can get the the "No Console Available" error. That's because the headless JDownloader will try to prompt the user for the MYJD details. Since it is being run in the background, instead of a regular terminal, there is no console to request this information in, so it throws this error. I'm hoping ultimately we can use this to verify MYJD details.
> > 
> > Does it still say it's active like this?
> > 
> > image
> > 
> > Because when I put in accurate details I stall out somewhere around here
> > 
> > image
> 
> I’d suggest use the default masteruser username and password for this. You should be able to find the _get_user_password function pretty easily. 

Unfortunately the swizzin user and password are never going to line up with the my.jdownloader.org details. I have now added in logic to verify the my jdownloader details are correct. Later on, I will add in a little bit more logic later to bypass inputting my.jdownloader.org account details.

Give it another spin and let me know how it goes. Interested to hear how it goes.

scripts/install/jdownloader.sh
```diff
@@ -0,0 +1,165 @@
...
+    fi
+
+    cat > "$JD_HOME/cfg/org.jdownloader.api.myjdownloader.MyJDownloaderSettings.json" << EOF
+{
+    "email" : "$MYJD_EMAIL",
+    "password" : "$MYJD_PASSWORD",
+    "devicename" : "$MYJD_DEVICENAME"
+}
+EOF
+}
```
brettpetch, line 54: this *could* be something in a functions file.

scripts/remove/jdownloader.sh
```diff
@@ -0,0 +1,25 @@
...
+. /etc/swizzin/sources/functions/utils
+
+# this will check for all swizzin users, and iterate and remove all JDownloader installations.
+users=("$(_get_user_list)")
+# The following line cannot follow SC2068 because it will cause the list to become a string.
+# shellcheck disable=SC2068
+for user in ${users[@]}; do
+    echo_progress_start "Removing JDownloader for $user..."
+    systemctl disable -q --now jdownloader@"$user"
+    # TODO: Do a check to see if they want to purge their JDownloader configurations.
```
brettpetch, line 16: we don't currently have purge functionality built into swizzin. It is being looked at for our v4 milestone as a box flag. If you did want to include this option, it might be a good idea to push into its own purge function ;)

scripts/install/jdownloader.sh
Aethaeran: Will likely be. Going to add a catch to inject MyJDownloader details into installations that don't have proper details yet.

Might have to move some of the MyJD verification into this function as well.

scripts/remove/jdownloader.sh
Aethaeran: Was assuming it would be purged on removal, or that I could add a simple yes/no here.
deluser purges all configurations, so I guess I just assumed a feature removal would as well.

Aethaeran: I've now added the yes/no prompts for configuration purging on removal.

At the very least I'd like to leave it this way until testing is thoroughly completed, then just perhaps move that logic and operation to a purge_jdownloader function.

scripts/install/jdownloader.sh
```diff
@@ -182,11 +182,12 @@ EOF
 _systemd
 
 # TODO: Add environment variable to bypass the following block. For unattended installs.
-if ask "Do you want to add ANY MyJDownloader account information for users?\nIt is required for them to access the web UI." N; then
-    NO_MYJD="false" # If no
-else
-    NO_MYJD="true" # If yes
-fi
+if [[ -n "${BYPASS_MYJDOWNLOADER}" ]]; then
```
flying-sausages, line 185: Let's stick to the `MYJD_` prefix please, it will be a lot more coherent in an env file down the road

PR #729, opened 2021-07-25
feat: add bullseye support

liaralabs:
## Description
Adds bullseye support
Updates mono installed version to 6.8 to match focal/bullseye
Reduces packages installed with python deps in deluge install

## Fixes issues: 
- bullseye doesn't work

## Proposed Changes:
- make bullseye work

## Change Categories
- Bug fix 
- New feature 

## Checklist
- [x] Testing was done

## Test scenarios
Deluge & qbit compile
panel/nginx works
all available rtorrent versions work

php untested (assume it will work as everything installs like focal)

### DIFF
--- a/sources/functions/mono
+++ b/sources/functions/mono
@@ -13,8 +13,8 @@ function mono_repo_setup() {
                 #gpg --export 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF > /etc/apt/trusted.gpg.d/mono-xamarin.gpg
             fi
         fi
-        if [[ ! $codename =~ ("buster"|"focal") ]]; then
-            echo "deb https://download.mono-project.com/repo/${distribution,,} ${codename}/snapshots/5.18/. main" > /etc/apt/sources.list.d/mono-xamarin.list
+        if [[ $codename =~ ^(stretch|bionic|buster)$ ]]; then
+            echo "deb https://download.mono-project.com/repo/${distribution,,} ${codename}/snapshots/6.8/. main" > /etc/apt/sources.list.d/mono-xamarin.list
         fi
         apt_update
     fi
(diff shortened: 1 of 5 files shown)


flying-sausages:
Do we want to do a full run-through of apps? Could skip ones which are similar like the Arrs, Python apps, bin-and-service ones, etc.

You could literally just install all of them in one go and then run `box test` to see what works and what doesn't.

sources/functions/mono
```diff
@@ -24,11 +24,11 @@ function mono_repo_update() {
     distribution=$(lsb_release -is)
     codename=$(lsb_release -cs)
     if [[ -f /etc/apt/sources.list.d/mono-xamarin.list ]]; then
-        if grep -q "5.18" /etc/apt/sources.list.d/mono-xamarin.list; then
+        if grep -q "6.8" /etc/apt/sources.list.d/mono-xamarin.list; then
             :
         else
-            echo "deb https://download.mono-project.com/repo/${distribution,,} ${codename}/snapshots/5.18/. main" > /etc/apt/sources.list.d/mono-xamarin.list
-            echo_log_only "Upgrading to mono 5.18 snapshot"
+            echo "deb https://download.mono-project.com/repo/${distribution,,} ${codename}/snapshots/6.8/. main" > /etc/apt/sources.list.d/mono-xamarin.list
```
flying-sausages, line 30: Should we update old installs with an updater?

liaralabs: You mean this updater?

https://github.com/swizzin/swizzin/blob/master/scripts/update/mono.sh#L9

flying-sausages: slick

flying-sausages:
I haven't checked this on a machine but white on black it looks good.

flying-sausages (approved):

PR #712, opened 2021-07-03
feat: readarr

flying-sausages:
- feat: readarr installer
- test: add test definition for readarr
- style: make the nginx config legible

## Description

Add readarr

## Fixes issues: 
N/A
## Proposed Changes:
New app with tests and all

## Change Categories

- New feature 

## Checklist

- [x] Branch was made off the `develop` branch and the PR is targetting the `develop` branch
- [x] Docs have been made OR are not necessary
    - PR link: https://github.com/liaralabs/docs.swizzin.ltd/pull/43
- [x] Changes to panel have been made OR are not necessary
    - PR link: https://github.com/swizzin/swizzin_dashboard/pull/24
- [x] Code is formatted [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)
- [x] Code conforms to project structure [(See more)](https://swizzin.ltd/dev/structure)
- [x] Shellcheck isn't screaming [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)
- [x] Prints to terminal are handled [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#printing-into-the-terminal)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] Testing was done
   - [x] Tests created or no new tests necessary
   - [x] Tests executed

### Architectures

|   			| `amd64` 	| `armhf` 	| `arm64` 	| Unspecified 	|
|--------		|-------- 	|-------- 	|-------- 	|----------		|
| Focal 		|			|			|	yup		|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|
| Raspbian  	|	⚫️		|			|	⚫️		|	⚫️			|

PR #734, opened 2021-08-12
feat: Autobrr

ludviglundgren:
## Description
Add support for `autobrr`, which is a modern alternative to autodl-irssi with built in support for new clients and more coming.
It's very much still work in progress and can't be considered stable yet. 
https://github.com/autobrr/autobrr

!! Pulled latest into the branch so unfortunately more stuff in it than only my things.

## Fixes issues: 
- Un-tracked issue: Add autobrr

## Proposed Changes:
- Add autobrr installer, remover and updater

## Change Categories
- New feature

## Checklist

    - PR link: https://github.com/liaralabs/docs.swizzin.ltd/pull/98
    - PR link: https://github.com/swizzin/swizzin_dashboard/pull/39
- [x] Testing was done
   - [x] Tests executed

## Test scenarios

Did manual testing only remote vm and local vm.

### Architectures

It has multi arch support thanks to golang but have not yet tested and verified other than amd64.

| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|	✅ 		|			|	 		|				|
| Stretch		|			|			|			|				|

### 🛠🛠 TODO

- [x] Multiuser support

### ❌❌ Currently failing

// ze0s


flying-sausages:
Heya, thanks for the PR, looks like a good fit for our project. If you do `git rebase origin/develop` and then `git push --force` you should get rid of the dupe commits.

I can also do that for you if you allowed maintainers to push to this branch. Let me know, and I can start reviewing.

How stable is the current build of `autobrr`? Do you have a roadmap of some sorts? FYI, We tend to delay merging applications which are in alpha/nightly stage (e.g. #712) and wait for a beta/development level of maturity; mostly because of support.

ludviglundgren:
Hello!

I added you as a maintainer so you should be able to do the cleanup. 

The latest release, v0.1.0 is stable enough to use, but more features are incoming and there might be breaking changes before a v1 is released. I'll try not to do it, but could happen. 

No roadmap yet, but some of the stuff on the todo list:

* add support for more trackers
* add support for Deluge
* add support for running commands on filtered (like autodl-irssi)
* add support to send to sonarr/radarr etc
* more advanced scheduling, rules like x per hour, max 2 downloads at a time and so on
* documentation! And this will probably keep people away for a while, but would be nice to have the installer in swizzin for easy deploy or releases.
* notifications
* automatic management like moving files, delete on X condition
* filter options specific to music, games and maybe more

brettpetch:
- [ ] Make Testing Suite (should be able to with the code i supplied in WS if it's MU)
- [ ] Multiuser
- [ ] Some kind of app auth?

ludviglundgren:
Fixed perms, and ran the suggested formatters.

ludviglundgren:
After a discussion with B we came to the conclusion that adding app level auth and making it per user/swizzin multuser will be the way to go. I'll start working on it and notify when it's done.

ludviglundgren:
Added 

> * [ ]  Make Testing Suite (should be able to with the code i supplied in WS if it's MU, if not you could just use the swizzdb var to get the owner user)
> * [ ]  Multiuser
> * [ ]  Some kind of app auth?
> * [ ]  Fix script perms (644 currently, should be 755)
> 
> Nice work otherwise though!

Fixed:

 * [x]  Multiuser ✅ 
 * [x]  Some kind of app auth? Built it's own auth system ✅ 
 * [x]  Fix script perms (644 currently, should be 755) ✅ 

Still unsure about the "Make testing suite". 
Still no docs, so if we can with with them that would be great.

flying-sausages:
> Still unsure about the "Make testing suite".

`box test` could probably guess most of what it needs to be default, try installing it and then run the command and then we can see what "exemptions" need to be made for it

brettpetch:
It’s an MU app, wouldn’t it need to be manual?

/b

>

flying-sausages:
Depends on where the auth is handled, i.e. whether it's one process running or one process per user

brettpetch:
> Depends on where the auth is handled, i.e. whether it's one process running or one process per user

It’s thru nginx now ;)

ludviglundgren:
Multiuser with per app install and auth per user. Some of the box test failed for it but not the nginx and curl test.

flying-sausages:
> Multiuser with per app install and auth per user. Some of the box test failed for it but not the nginx and curl test.

Cool, I can take a look once the deployment method is finalised

ludviglundgren:
@brettpetch helped me with the setup and if he doesn't have anything to add then it's done from my side.

flying-sausages:
> @brettpetch helped me with the setup and if he doesn't have anything to add then it's done from my side.

oh shit I might have just possibly accidentally pushed over some of your commits while i was rebasing on top of develop

flying-sausages:
>  Still no docs, so if we can with with them that would be great.

This is another docs app addition PR from earlier https://github.com/liaralabs/docs.swizzin.ltd/pull/81/files

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
+#!/bin/bash
+# autobrr installer
+# ludviglundgren 2021 for Swizzin
+
+#shellcheck source=sources/functions/utils
+. /etc/swizzin/sources/functions/utils
+
+app_reqs=("curl" "sqlite3")
+apt_install "${app_reqs[@]}"
```
flying-sausages, line 9: I see where you pulled this from, but that's an "arr" format, tl;dr:
```suggestion
apt_install curl sqlite3
```

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+    user=$1
+    _autobrr_user_config ${user}
+    if [[ -f /install/.nginx.lock ]]; then
+        echo_progress_start "Configuring nginx"
+        bash /etc/swizzin/scripts/nginx/autobrr.sh
+        systemctl reload nginx
+        echo_progress_done
+    fi
+    exit 0
+fi
```
flying-sausages, line 23: This whole section might have to go under the function definitions as `_autobrr_user_config` will not be defined at this point

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+    if ! curl "$latest" -L -o "/tmp/autobrr.tar.gz" >> "$log" 2>&1; then
+        echo_error "Download failed, exiting"
+        exit 1
+    fi
+    echo_progress_done "Archive downloaded"
+
+    echo_progress_start "Extracting archive"
+
+    # the archive contains both autobrr and autobrrctl to easily setup the user
+    tar xfv "/tmp/autobrr.tar.gz" --directory /usr/bin/ >> "$log" 2>&1 || {
```
flying-sausages, line 53: no need to ensure perms to `+x`? if the perms are stored in the archive then it won't be an issue, just curious

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+    echo_progress_start "Configuring autobrr"
+
+    # get random available port
+    port=$(port 10000 12000)
+
+    # generate a sessionSecret
+    sessionSecret="$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c16)"
+
+    if [ ! -d "/home/$user/.config/autobrr/" ]; then
+        mkdir -p "/home/$user/.config/autobrr/"
```
flying-sausages, line 71: Just in case this is the first app a user installs (it starting with `a`, high chance), gotta make sure that `~/.config` is not owned by `root`
```suggestion
        mkdir -p "/home/$user/.config/autobrr/"
        chown "$user": /home/"$user"/.config
```

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+    fi
+    chown -R "$user": "/home/$user/.config/autobrr"
+
+    cat > "/home/$user/.config/autobrr/config.toml" << CFG
```
flying-sausages, line 75: Not a requirement or anything but if the `#` lines are not necessary, I'd just wipe them out and add a link to the full content of the default or some wiki page where users can find more.

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+    # get random available port
+    port=$(port 10000 12000)
+
+    # generate a sessionSecret
+    sessionSecret="$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c16)"
+
+    if [ ! -d "/home/$user/.config/autobrr/" ]; then
+        mkdir -p "/home/$user/.config/autobrr/"
+    fi
+    chown -R "$user": "/home/$user/.config/autobrr"
```
flying-sausages, line 73: L113 dupe, we only need the last one of these that will run

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+    # use autobrrctl to add user into database
+    echo_progress_start "Add user"
+
+    for user in "${users[@]}"; do
+        echo_log_only "Adding user $user"
+        pass=$(_get_user_password "$user")
+
+        # the password needs to be created with argon2 so we use autobrrctl to create the user
+        # using sqlite3 directly was not an option
+        echo -n "$pass" | autobrrctl --config "/home/$user/.config/autobrr" create-user "$user"
```
flying-sausages, line 129: pipe pipe pipe
```suggestion
        # the password needs to be created with argon2 so we use autobrrctl to create the user
        # using sqlite3 directly was not an option
        _get_user_password "$user" | autobrrctl --config "/home/$user/.config/autobrr" create-user "$user"
```

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+    done
+}
+
+_systemd_autobrr() {
+    if [[ ! -f /etc/systemd/system/autobrr@.service ]]; then
```
flying-sausages, line 134: I'd just _always_ overwrite the systemd service if this is being installed. Otherwise some edge case could keep on re-using an out of date service file. i.e. I'd remove this `if fi`

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+After=syslog.target network.target
+
+[Service]
+Type=$type
+User=%i
+Group=%i
+ExecStart=/usr/bin/autobrr --config=/home/%i/.config/autobrr/
+# TimeoutStopSec=20
+# KillMode=process
+# Restart=on-failure
```
flying-sausages, line 153: Are these still to decide? Otherwise I'd yank 'em out or use them

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+
+if [[ -n $1 ]]; then
+    user=$1
+    _autobrr_user_config ${user}
+    if [[ -f /install/.nginx.lock ]]; then
+        echo_progress_start "Configuring nginx"
+        bash /etc/swizzin/scripts/nginx/autobrr.sh
+        systemctl reload nginx
+        echo_progress_done
+    fi
```
flying-sausages, line 21: The config function does not start the service it seems
```suggestion
    fi
    systemctl enable -q --now autobrr@${user} 2>&1 | tee -a $log
```

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+        bash /etc/swizzin/scripts/nginx/autobrr.sh
+        systemctl reload nginx >> $log 2>&1
+        echo_progress_done "Nginx configured"
+    fi
+}
+
+_install_autobrr
+
+_systemd_autobrr
+_systemd_autobrr_enable_for_user
```
flying-sausages, line 183: Might as well merge these into one

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+_autobrr_add_user
```
flying-sausages, line 184: I'd add an `s` to the end to make this a bit more different to `_autobr_config_user` (seeing as it also processes more than one user)
```suggestion
_autobrr_add_users
```

scripts/nginx/autobrr.sh
```diff
@@ -0,0 +1,36 @@
...
+# ludviglundgren 2021 for Swizzin
+
+#shellcheck source=sources/functions/utils
+. /etc/swizzin/sources/functions/utils
+users=($(_get_user_list))
+
+if [[ ! -f /etc/nginx/apps/autobrr.conf ]]; then
+
+    cat > /etc/nginx/apps/autobrr.conf << 'AUTOBRR'
+location /brr {
```
flying-sausages, line 12: lol :p are you sure `/brr` 301ing to `/autobrr/` won't be a problem with double-slashes? I never know myself about this when it comes to these nginx things

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+    fi
+    chown -R "$user": "/home/$user/.config/autobrr"
+
+    cat > "/home/$user/.config/autobrr/config.toml" << CFG
+# config.toml
+# Hostname / IP
+#
+# Default: "localhost"
+#
+host = "127.0.0.1"
```
flying-sausages, line 81: Is it ok for this to be on `127.0.0.1` if someone does not use `nginx`? We usually allow this to be `*` or similar by default, and then in the event of `nginx` we change it to `127...1` only. We _did_ see some issues with this recently in the arrs, @liaralabs any advice on this? Some call to make to also apply going forward?

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+#
+# Optional
+#
+baseUrl = "/autobrr/"
+# autobrr logs file
+# If not defined, logs to stdout
+#
+# Optional
+#
+#logPath = "log/autobrr.log"
```
flying-sausages, line 99: This would be nice to enable, `~/.config/autobrr/log.txt` or similar would be good enough. Especially if the output is set to `DEBUG`, that could fill up the `journalctl` fast. Had that happen with transmission before (spamming stdout while in service) and it just _kills_ performance of all `systemctl` etc. commands

scripts/upgrade/autobrr.sh
```diff
@@ -0,0 +1,60 @@
...
+
+#shellcheck source=sources/functions/utils
+. /etc/swizzin/sources/functions/utils
+
+if [[ ! -f /install/.autobrr.lock ]]; then
+    echo_error "autobrr doesn't appear to be installed!"
+    exit 1
+fi
+
+_upgrade_autobrr() {
```
flying-sausages, line 13: I noticed this code is duplicate from the install script, this could be de-duped by defining the function once in `sources/functions/autobrr` and then using it both in the install script as well as here

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,189 @@
...
+
+#shellcheck source=sources/functions/utils
+. /etc/swizzin/sources/functions/utils
+
+app_reqs=("curl" "sqlite3")
+apt_install "${app_reqs[@]}"
+
+users=($(_get_user_list))
+
+if [[ -n $1 ]]; then
```
flying-sausages, line 13: You will need to modify the useradd function inside `script/box` in order for it to trigger this. Also, please make a case for the password changing mechanism in the same file, so that the passwords are kept in sync everywhere when they're changed

flying-sausages (changes requested):
That's a good-looking PR! First read-through, these are some things that popped out at me. Gonna give this a run on a fresh machine tomorrow.

scripts/install/autobrr.sh
flying-sausages: (this will also need the actual function def name to be changed, mind you)

scripts/nginx/autobrr.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+# autobrr nginx conf
+# ludviglundgren 2021 for Swizzin
+
+#shellcheck source=sources/functions/utils
+. /etc/swizzin/sources/functions/utils
+users=($(_get_user_list))
+
+if [[ ! -f /etc/nginx/apps/autobrr.conf ]]; then
```
flying-sausages, line 9: I would also not check the conditional here, just write it whether it exists or not

scripts/install/autobrr.sh
ludviglundgren: I'd like to keep it in the config.

scripts/install/autobrr.sh
ludviglundgren: The thought before was to put it behind basic auth on the reverse proxy, but now that the app have auth we could potentially expose it on 0.0.0.0.

scripts/install/autobrr.sh
ludviglundgren: Interesting! Currently it outputs to both the console and logfile if set. It has mechanisms to rotate logs as well so it doesn't fill up the disk. Will have to look into writing only to log file if set, and nothing to console then.

scripts/install/autobrr.sh
ludviglundgren: The packager handles all that so the binaries will always be executable.

scripts/install/autobrr.sh
ludviglundgren: Great improvement!

scripts/install/autobrr.sh
ludviglundgren: Yeah I'm no super familiar with all systemd options, so this was copied from another service. I have removed them now since I'm not sure if they are needed or not.

scripts/install/autobrr.sh
ludviglundgren: Renamed and done.

scripts/nginx/autobrr.sh
ludviglundgren: Maybe not. Removed it 👍

scripts/install/autobrr.sh
ludviglundgren: Would that be added here? https://github.com/swizzin/swizzin/blob/6f07eb7f1ed54ccea217d01d45c658b1fbc8b1be/scripts/box#L306-L331

Re the password changing mech, I'll add a function for that to `autobrrctl`.

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,146 @@
...
+_autobrr_create_users() {
+    # use autobrrctl to add user into database
+    echo_progress_start "Add user"
+
+    for user in "${users[@]}"; do
+        echo_log_only "Adding user $user"
+
+        # the password needs to be created with argon2 so we use autobrrctl to create the user
+        # using sqlite3 directly was not an option
+        _get_user_password "$user" | autobrrctl --config "/home/$user/.config/autobrr" create-user "$user"
```
flying-sausages, line 94: I think `autobrrctl` needs to have the full path specified here btw, it's not showing up in my `path`

afbeelding

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,146 @@
...
+_nginx_autobrr() {
+    if [[ -f /install/.nginx.lock ]]; then
+        echo_progress_start "Configuring nginx"
+        bash /etc/swizzin/scripts/nginx/autobrr.sh
+        systemctl reload nginx >> $log 2>&1
+        echo_progress_done "Nginx configured"
+    fi
+}
+
+autobrr_download_latest
```
flying-sausages, line 138: The function still needs to be pulled in
```suggestion
#shellcheck source=sources/functions/autobrr
. /etc/swizzin/sources/functions/autobrr
autobrr_download_latest
```

scripts/upgrade/autobrr.sh
```diff
@@ -0,0 +1,24 @@
...
+
+_restart_autobrr() {
+    for user in "${users[@]}"; do
+        # restart autobrr
+        systemctl try-restart "autobrr@${user}"
+    done
+    echo_progress_done "Service restarted"
+}
+
+autobrr_download_latest
```
flying-sausages, line 21: Function missing here
```suggestion
#shellcheck source=sources/functions/autobrr
. /etc/swizzin/sources/functions/autobrr
autobrr_download_latest
```

scripts/install/autobrr.sh
flying-sausages: Yup, that's the one for `box adduser`. `box chpasswd` and `box deluser` work in a similar-ish fashion

scripts/install/autobrr.sh
flying-sausages: Yeah, I think that would be optimal. As long as it gets changed to `127.0.0.7` from the script for `nginx`

scripts/install/autobrr.sh
ludviglundgren: It only seems to be needed to add in the list at 306 right? Because the line 317 would trigger it, so extra check needed?

flying-sausages: Yup! If it has no special procedure, all it needs is to be in the list

scripts/install/autobrr.sh
ludviglundgren: I set it to log to log to `/home/$user/.config/autobrr/logs/autobrr.log` and will make some changes to only have some stuff output to console/stdout. But that will come with an update to the app :)

flying-sausages (approved):

flying-sausages:
@brettpetch @liaralabs can one of you please do another full check that this performs as it should? I would say this is good to merge to develop

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,134 @@
...
+            exit 1
+        }
+
+        # Only run the nginx hook if a new user is being added, otherwise it will be ran for all users at the end of the installer
+        if [[ ${#users[@]} -eq 1 ]]; then
+            if [[ -f /install/.nginx.lock ]]; then
+                bash /etc/swizzin/scripts/nginx/autobrr.sh
+                systemctl reload nginx
+            fi
+        fi
```
liaralabs, line 109: This snippet will fire if the master user is the only user on the seedbox when autobrr is installed.

It's a bit more clunky, but a variable could be set between lines 120-124 that indicates that `adduser` is the current environment rather than `install`. That is, assuming that double nginx setup is an issue; however, in theory it should not be

scripts/install/autobrr.sh
```diff
@@ -0,0 +1,134 @@
...
+#
+# Options: "ERROR", "DEBUG", "INFO", "WARN"
+#
+logLevel = "DEBUG"
+
+# Session secret
+#
+sessionSecret = "${sessionSecret}"
+CFG
+        chown -R "$user": "/home/$user/.config/autobrr"
```
liaralabs, line 96: Run this command after creating the user (line 98) or suffer read-only root owned db :smile:

scripts/install/autobrr.sh
flying-sausages: resolved, logic moved into the main's `if $1` statement so it's outside the loop

PR #719, opened 2021-07-13
fix(arrs): Ensure sonarr and radarr have populated swizdb owners

flying-sausages:
## Description
Scope is to ensure that the `swizdb` reliably has the owners from the arr apps

## Fixes issues: 
- N/A

## Proposed Changes:

- fix: ensure sonarr owner is recorded
- fix: remove vars on uninstall
- style: fix echos
- fix: remove double apt update because mono_repo_setup has one
- fix: ensure .config gets right ownership if it exists
- fix: record radarr owner properly on install
- fix: purge config explicitly for radarr
- style: echos and formatting
- fix: rewrite variable recording for radarr updater
- style: echos, formatting, sanity

## Change Categories

- Bug fix 

## Checklist

- [x] Branch was made off the `develop` branch and the PR is targetting the `develop` branch
- [x] Docs have been made OR are not necessary
    - PR link: https://github.com/liaralabs/docs.swizzin.ltd/pull/89
- [x] Changes to panel have been made OR are not necessary
    - PR link: 
- [x] Code is formatted [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)
- [x] Code conforms to project structure [(See more)](https://swizzin.ltd/dev/structure)
- [x] Shellcheck isn't screaming [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#editor-plugins-and-tooling)
- [x] Prints to terminal are handled [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#printing-into-the-terminal)
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] Testing was done
   - [ ] Tests created or no new tests necessary
   - [ ] Tests executed

### Architectures

|   			| `amd64` 	| `armhf` 	| `arm64` 	| Unspecified 	|
|--------		|-------- 	|-------- 	|-------- 	|----------		|
| Focal 		|			|			|	WIP		|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|
| Raspbian  	|	⚫️		|			|	⚫️		|	⚫️			|

PR #741, opened 2021-08-17
update(sonarr): use tar ball for installation to prevent unsupported repo issues

liaralabs:
## Description
Moves sonarr to utilizing the tar ball installation as bullseye does not have a repo to install from, yet it shouldn't matter

## Fixes issues: 
- Bullseye cannot install Sonarr

## Change Categories

- Bug fix 

## Checklist

    - PR link: 
    - PR link: 
- [x] Testing was done
   - [ ] Tests executed

### ✅❎ Passed
amd64 on buster:

- Fresh install
- Update install (apt to tarball) -- data is moved to .config/Sonarr to allow purging sonarr to remove it from `dpkg -l`
- Removal

amd64 on bullseye:
- Fresh install
- Remove

### 🛠🛠 TODO
- Test ARM

### ❌❌ Currently failing

- None

### DIFF
--- a/scripts/update/sonarr.sh
+++ b/scripts/update/sonarr.sh
@@ -34,6 +34,7 @@ fi
 if [[ -f /install/.sonarr.lock ]] && dpkg -l | grep sonarr | grep ^ii > /dev/null 2>&1; then
     echo_info "Migrating Sonarr away from apt management"
     isActive=$(systemctl is-active sonarr)
+    isEnabled=$(systemctl is-enabled sonarr)
     cp -a /usr/lib/sonarr/bin /opt/Sonarr
     cp /usr/lib/systemd/system/sonarr.service /etc/systemd/system
     user=$(grep User= /etc/systemd/system/sonarr.service | cut -d= -f2)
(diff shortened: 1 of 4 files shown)


bakerboy448:
just for posterity

> bullseye does not have a repo to install from, yet it shouldn't matter

Believe focal's repo should work fine as a fallback.

> Moves sonarr to utilizing the tar ball installation

Officially not supported / advanced user yadda yadda.  but given the deps and everything will be handled by swizzin. should be fine.

flying-sausages:
@liaralabs what about applying #719 first?

scripts/update/sonarr.sh
```diff
@@ -42,3 +31,65 @@ if [[ -f /install/.sonarrv3.lock ]]; then
...
+    rm -f /etc/apt/sources.list.d/sonarr.list*
+    systemctl daemon-reload
+
+    #Restart sonarr because apt-removal stops it
+    if [[ $isActive == "active" ]]; then
+        systemctl start sonarr
+    fi
+
+    #Update broken symlink to old service
+    if systemctl is-enabled sonarr > /dev/null 2>&1; then
```
flying-sausages, line 91: My service was running but disabled after the update

liaralabs:
> Believe focal's repo should work fine as a fallback.

Not a huge fan of this workaround.

>  but given the deps and everything will be handled by swizzin

deps were pulled directly from the sonarr deb package, so I believe it should be okay. If required deps somehow change before the dotnet version finally ships then we can associate that into the updater. However, I chose to use the exact deps rather than install `mono-complete` as the `mono-complete` package is massive and it's clear that sonarr doesn't require the full base to function.

scripts/update/sonarr.sh
liaralabs: Updated to check the variable before hand and apply the fix based on previously reported status.

flying-sausages (approved):

PR #743, opened 2021-08-19
Feat: LazyLibrarian

Aethaeran:
## Description

Add requested feature lazy librarian

## Fixes issues: 
- Issue #nr...
- Un-tracked issue...

## Proposed Changes:
- Changed xyz...

## Change Categories

- Bug fix 
- New feature 
- Breaking change 
- Change in `functions` 
- Project structure 
- Contribution flow 

## Checklist

    - PR link: 
    - PR link: 
- [ ] Testing was done
   - [ ] Tests executed

### Architectures

| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|


flying-sausages:
Yo I'm moving this into Draft for now, please flip it back when it's ready to review and then request a review from me. Thanks!

Aethaeran:
Ready for review.

scripts/install/lazylibrarian.sh
```diff
@@ -0,0 +1,162 @@
...
+# https://lazylibrarian.gitlab.io/api/
+
+##########################################################################
+# Import Sources
+##########################################################################
+
+. /etc/swizzin/sources/functions/color_echo
+. /etc/swizzin/sources/functions/pyenv
+. /etc/swizzin/sources/functions/users
+. /etc/swizzin/sources/functions/utils
```
flying-sausages, line 28: These are missing the `#shellcheck source=` directives, e.g.

```bash
    #shellcheck source=sources/functions/mono
    . /etc/swizzin/sources/functions/mono
```

scripts/install/lazylibrarian.sh
```diff
@@ -0,0 +1,162 @@
...
+    mkdir -p "$venv_dir"
+    python3_venv "$master" "$app_name"
+    echo_progress_done
+
+    echo_progress_start "Installing python dependencies to venv"
+    # shellcheck disable=2154           # log variable is inherited from box itself.
+    "$venv_dir/bin/pip" install --upgrade pip >>"${log}" 2>&1 # Upgrade PIP
+    PIP='apprise pyOpenSSL'
+    # shellcheck disable=2086           # We want the $PIP variable to expand. So 2086's warning is invalid here.
+    "$venv_dir/bin/pip" install $PIP >>"${log}" 2>&1
```
flying-sausages, line 59: all cap variables are "reserved" for environment variables, please use something like `pipreqs` or something. Could also move it to the very top

OR, just skip the variable altogether if it's used just once

scripts/install/lazylibrarian.sh
```diff
@@ -0,0 +1,162 @@
...
+    chown -R "$master": "$venv_dir"
+    echo_progress_done
+    # TODO: https://lazylibrarian.gitlab.io/config_rtorrent/ says we would require libtorrent for rtorrent compatability
```
flying-sausages, line 62: @liaralabs what to do about this? Seems like a pretty strong dependency for what I assume 80% of users will be doing. If we leave this up to the users, it will get messy as hell

scripts/install/lazylibrarian.sh
```diff
@@ -0,0 +1,162 @@
...
+    mkdir -p "$data_dir"
+    #    Git clone/extract LL wherever you like
+    # shellcheck disable=2154           # log variable is inherited from box itself.
+    git clone "https://gitlab.com/LazyLibrarian/LazyLibrarian.git" "$app_dir" >>"$log" 2>&1
+    if [[ ! -e "$app_dir/LazyLibrarian.py" ]]; then
+        ecoh_error "Git clone unsuccessful. Try running the box install again."
+        exit 1
+    fi
+    chown -R "$master": "$app_dir" # Change owner\group recursively for new dirs.
+    chown -R "$master": "$data_dir"
```
flying-sausages, line 77: Need to add a non-recusrive chown on `~/.config` so it won't be owned by root

scripts/install/lazylibrarian.sh
```diff
@@ -0,0 +1,162 @@
...
+Description=LazyLibrarian
+After=network.target
+
+[Service]
+User=$master
+Group=$master
+Type=simple
+ExecStart=$venv_dir/bin/python \
+$app_dir/LazyLibrarian.py \
+--datadir $data_dir
```
flying-sausages, line 118: I would make this one line and also maybe with `--nolaunch`

flying-sausages: Would we want to also make the pidfile just in case?

scripts/remove/lazylibrarian.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
```
flying-sausages, line 1: In general, we don't do `echo_`s during removers, could just switch them to `echo_log_only`

scripts/install/lazylibrarian.sh
```diff
@@ -0,0 +1,162 @@
...
+. /etc/swizzin/sources/functions/utils
+
+##########################################################################
+# Variables
+##########################################################################
+
+app_name="lazylibrarian"
+pretty_name="LazyLibrarian"
+default_port="5299"
+master=$(_get_master_username)
```
flying-sausages, line 37: I would suggest making this an environment variable, you can check #719 how that looks. You just need to paste the "logic" for retrieving that into the master variable. Might want tot rename that to "user" as well.

scripts/install/lazylibrarian.sh
flying-sausages: If #746 is merged first, maybe no need

flying-sausages (commented):
Minor things here and there, will try running the installer later on

scripts/install/lazylibrarian.sh
Aethaeran: Can change. I copied this variable naming convention from another install script. Would you like me to make a quick PR changing it in other scripts as well?

scripts/install/lazylibrarian.sh
Aethaeran: Either --nolaunch or --daemon. I assume --daemon implies --nolaunch. Will fiddle with it a bit today.

PIDFILE makes sense to me. I left all those flags there so you would tell me which might be handy. ;-)

Can make it a one liner, but in practice they are identical. The service generated has a single line regardless if it's multiline here or not. I just find the code slightly easier to read this way. Up to you.

scripts/install/lazylibrarian.sh
Aethaeran: Most of them are imported by box already anyways. I just put them there so I would remember to source them to CLI when testing individual blocks.

Will add the "shellcheck sources" to the ones that remain after verifying which are necessary or not.

scripts/remove/lazylibrarian.sh
Aethaeran: Sure. I'll put some TODOs in to change them to log_only near PR end. Would like the output to be sent to CLI for the time being.

scripts/install/lazylibrarian.sh
Aethaeran: My two cents? 

If rTorrent needs libtorrent for magnet links. Then libtorrent should be added during rTorrent's install. 

However, if rTorrent simply doesn't have magnet functionality on its own. Then we should have a lock check in both installers that will install libtorrent if the other app is already installed.

I don't know about you, but I like my torrent download client to be able to accept magnet links. 😅

scripts/install/lazylibrarian.sh
Aethaeran: Done.
PyCharm's shellcheck never says anything about "shellcheck source". So I never would have caught this on my own. Thanks a bunch.

scripts/install/lazylibrarian.sh
Aethaeran: Changed `PIP` to `pip_reqs`. Moved its declaration up with the rest of the variables.

scripts/install/lazylibrarian.sh
Aethaeran: Added it in anyways. Couldn't hurt.

scripts/install/lazylibrarian.sh
Aethaeran: I like this idea because then people could optionally install it to a different user using the variable. Smart cookie.

scripts/install/lazylibrarian.sh
Aethaeran: Alright. Looks like all three will be needed.
`--nolaunch` because `--daemon` doesn't trigger `--nolaunch`
`--daemon` because `--pidfile` does nothing without it.
`--pidfile` for an easy outside reference to which PID the daemon is run on.

scripts/remove/lazylibrarian.sh
flying-sausages: Pro tip: run `tail -f /root/logs/swizzin.log &` and hit enter. You'll get the full log including log only and forwarded output into your terminal as the script runs.

scripts/install/lazylibrarian.sh
flying-sausages: should be done now

scripts/install/lazylibrarian.sh
Aethaeran: Done

PR #744, opened 2021-08-19
fix(setup): arm64, improvements to flags like `--test` and `--env`

flying-sausages:
## Description

This was preventing me from unattended installs on amd64 which is well supported now

## Fixes issues:
- fix: add local dev environment to gitignore
- feat: mark arm64 as supported
- fix: change `--test` flag in setup.sh
- fix: run tests when `.test.lock` is present

## Proposed Changes:
- mark amd64 as supported and stop telling people they can't make github issues

## Change Categories

- New feature 

## Checklist

    - PR link: https://github.com/liaralabs/docs.swizzin.ltd/pull/97
    - PR link: 
- [x] Testing was done
   - [x] Tests executed

### Architectures

| Focal 		|			|			|	yup		|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -60,19 +60,16 @@ function _source_setup() {
 _source_setup "$@"
 
 function _arch_check() {
-    if [[ ! $(uname -m) == "x86_64" ]]; then
+    if [[ ! $(uname -m) =~ ("x86_64"|"aarch64") ]]; then
         echo_warn "$(_os_arch) detected!"
-        if [[ $(_os_arch) = "arm64" ]]; then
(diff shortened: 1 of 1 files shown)


liaralabs:
I actually feel like we should be running the os and arch check before we do anything at all. I don't think we need to clone the project to tell people that they can't use the setup script in the current state. I know we went through the trouble of making sure the pretty echos work for these prompts, but I would rather just have the script exit early.

I'm also noticing that a lot of the newly added functions got added in a way that breaks the flow the setup originally had -- the newer functions are added and then immediately run; whereas the original code added all the functions and then ran them separately at the bottom. This is how the os check ended up a bit buried I think.

flying-sausages:
I'm splitting this whole thing into separate PRs as the scope of this is creeping

PR #795, opened 2021-11-15
(fix) use most recent kepubify not a very old version

bakerboy448:
## Fixes issues: 
- Un-tracked issue...

## Proposed Changes:
get most recent kepubify

## Change Categories

- Bug fix 

## Checklist

- [X] Branch was made off the `develop` branch and the PR is targetting the `develop` branch
    - PR link: 
- [X] Changes to panel have been made OR are not necessary
    - PR link: 
- [X] Code conforms to project structure [(See more)](https://swizzin.ltd/dev/structure)
- [X] Testing was done
   - [X] Tests created or no new tests necessary
   - [X] Tests executed

### Architectures

| Focal 		|	✅		|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|


scripts/install/calibreweb.sh
```diff
@@ -89,7 +89,11 @@ function _install_calibreweb() {
 
 _install_kepubify() {
     echo_progress_start "Installing kepubify"
-    wget -q "https://github.com/pgaskin/kepubify/releases/download/v3.1.2/kepubify-linux-64bit" -O /tmp/kepubify >> $log 2>&1
+    kepdlurl=$(curl -s https://api.github.com/repos/pgaskin/kepubify/releases/latest | grep "browser_download_url" | grep "kepubify-linux-64bit" | head -1 | cut -d\" -f 4) || {
+        echo_error "Failed to query github"
+        exit 1
+    }
+    wget -q "$kepdlurl" -O /tmp/kepubify >> $log 2>&1
```
userdocs, line 96: I would suggest this instead

```bash
_install_kepubify() {
	echo_progress_start "Installing kepubify"

	case "$(_os_arch)" in
		"amd64")
			kepubify_arch="64bit"
			;;
		"arm64")
			kepubify_arch="arm64"
			;;
		"armhf")
			kepubify_arch="arm"
			;;
		*)
			echo_error "$(_os_arch) not supported by this app"
			exit 1
			;;
	esac

	kepubify_url="https://github.com/pgaskin/kepubify/releases/latest/download/kepubify-linux-${kepubify_arch}"
	kepubify_url_status="$(curl -sLo /dev/null -w "%{http_code}" "${kepubify_url}")"

	if [[ "${kepubify_url_status}" -eq '200' ]]; then
		wget "${kepubify_url}" -O /usr/local/bin/kepubify
		chmod 755 /usr/local/bin/kepubify
	else
		echo_error "kepubify github release URL status: ${kepubify_url_status}"
		exit 1
	fi
	echo_progress_done
} 2>> "${log}"
```

scripts/install/calibreweb.sh
```diff
@@ -88,13 +88,36 @@ function _install_calibreweb() {
...
+
+	if [[ "${kepubify_url_status}" -eq '200' ]]; then
+		wget "${kepubify_url}" -O /usr/local/bin/kepubify
+		chmod 755 /usr/local/bin/kepubify
+	else
+		echo_error "kepubify github release URL status: ${kepubify_url_status}"
+		exit 1
+	fi
+	echo_progress_done
+} 2>> "${log}"
```
userdocs, line 120: The reason this will work is because the release [artifacts do not include version info](https://github.com/pgaskin/kepubify/releases) so this url will always provide the latest release asset for the relevant arch.

https://github.com/pgaskin/kepubify/releases/latest/download/kepubify-linux-64bit
https://github.com/pgaskin/kepubify/releases/latest/download/kepubify-linux-arm64
https://github.com/pgaskin/kepubify/releases/latest/download/kepubify-linux-arm

We are ignoring 32bit

So using the case statement we can create a url to always:

- pull the latest release of kepubify
- always get the matching arch for supported systems

So this fixes a few things

- This fixes the issue of having an old version hardlinked
- It fixes the single arch `64bit` configuration by supporting arm64 and arm
- Check the URL is valid for a `200` return code.
- errors out if the url fails.

I think this fixes the main problem and well as future proofs it.

bakerboy448: makes sense to me!

bakerboy448:
This will also need an updater to fix the existing hardcoded installs

userdocs:
and the removal of the binaries.

the function here can basically be the updater as well. i don't think we need to test versions or be complex.

sources/functions/kepubify
```diff
@@ -0,0 +1,66 @@
...
+        "armhf")
+            kepubify_arch="arm"
+            ;;
+        *)
+            echo_error "$(_os_arch) not supported by this app"
+            exit 1
+            ;;
+    esac
+
+    kepubify_url="https://github.com/pgaskin/kepubify/releases/latest/download"
```
brettpetch, line 44: Can we move this to use the new `github_latest_version` instead before merge? I think that now that we have this as a function we should just use that.

brettpetch:
@bakerboy448 is there anything to be changed here?

sources/functions/kepubify
bakerboy448: Probably, do the thing ;)

bakerboy448 (commented):
Housekeeping and maybe a testing? but otherwise probably fine

Issue #811, opened 2022-01-11
[BUG] Rutorrent fails with Throttle plugin(most likely due to a hardcoded rutorrent commit)


tomerblecher:
### What happened?

Throttle plugin(and I suspect some others as well such as ExtRatio) causes the Rutorrent to fail.
After opening a ticket to Rutorrent themselves,  https://github.com/Novik/ruTorrent/issues/2259 - they responded that if I'm not being updated with master, I should use the plugins from the latest stable version(v3.10).

That made me dig into the installation script where I found that you indeed checkout to a specific commit(I also checked your latest commits and saw that it was added a week~ ago https://github.com/swizzin/swizzin/commit/6757060eae0649fb8d9a7da941325f3e198293ab), but the rtx script that is being used to download the plugins, uses the latest revision with svn.

Potential fix would be to hardcode the version in the rtx script as well.

Steps to reproduce:

1. Install package rutorrent
2. instsall plugin Throttle
3. Nginx will throw errors while loading the plugin which will cause the GUI to fail load.

Expected results:
No bugs :)

### Swizzin commit

9ef0807

### What OS are you using?

Ubuntu 18.04 (Xenial)

### What architecture is your OS?

amd64

### Relevant logs and output

```plaintext
Nginx:
PHP message: PHP Fatal error: Uncaught Error: Class 'FileUtil' not found in /srv/rutorrent/plugins/throttle/throttle.php:4
Stack trace:
#0 /srv/rutorrent/plugins/throttle/init.php(2): require_once()
#1 /srv/rutorrent/php/getplugins.php(495): require_once('/srv/rutorrent/...')
#2 {main}
```

chiphead2332:
rutorrent 4.0 beta was released a couple of days ago, upgrading everything (including plugins) to the new version should work.

liaralabs:
This is an okay solution if you don't use AutoDL, but the stem of this problem is because ruTorrent was pinned to a version that keeps the autodl plugin working. Anything past the pinned commit will break the plugin. There's an open PR with no feedback from the author at the moment. Unclear if/when it will get merged.

https://github.com/autodl-community/autodl-rutorrent/pull/58

stickz:
How would you feel about forking the project, if we don't hear back from the author in anther week? ruTorrent 4.0 beta has some very substantial optimizations and bug fixes. To name a couple:

1. A race condition was resolved which could cause the WebUI settings to not save properly.
[...]

PR #816, opened 2022-01-23
fix(autodl/rutorrent): update ruTorrent and pull autodl upstream into swizzin org w/ fixes

liaralabs:
## Description
Our ruTorrent "quick fix" had reverberating implications on being able to install plugins from head. Basically you can't and installing little things like throttle after the fact is essentially impossible without massive work arounds. This is all due to autodl ruTorrent plugin being out of date.

Since we don't support ruTorrent upgrades, the only official way to "fix" this on a current installation is

```
box remove rutorrent
box install rutorrent
```

## Fixes issues: 
- #811 

## Proposed Changes:
- autodl-irssi plugin upstream has been pulled into the swizzin org with fixes applied for ruTorrent 4.0
- Unpin ruTorrent version
- Deprecate broken plugins
- Fix filemanager plugins to use new version since the are broken/deprecated
- Fix nginx fancyindex to allow php with split paths

## Change Categories

- Bug fix 

## Checklist

- [ ] Docs have been made OR are not necessary (some ruTorrent docs could be good here)
    - PR link: 
- [x] Testing was done

## Test scenarios
Only tested under Bullseye could potentially use a bit more coverage before merging

### DIFF
--- a/scripts/rtx
+++ b/scripts/rtx
@@ -109,12 +109,12 @@ function _itplugs() {
             filemanager-share)
                 git clone https://github.com/nelu/rutorrent-filemanager-share.git filemanager-share > ${log} 2>&1
                 ln -s /srv/rutorrent/plugins/filemanager-share/share.php /srv/fancyindex/share.php > /dev/null 2>&1
-                sed -i 's/$downloadpath .*/$downloadpath = $_SERVER['"'HTTP_HOST'"'] . '"'\/fancyindex\/share.php'"';/g' /srv/rutorrent/plugins/filemanager-share/conf.php
+                sed -i 's/$downloadpath .*/$downloadpath = '"'https:\/\/'"'.$_SERVER['"'HTTP_HOST'"'] . '"'\/fancyindex\/share.php'"';/g' /srv/rutorrent/plugins/filemanager-share/conf.php
                 ;;
             filemanager-media)
                 git clone https://github.com/nelu/rutorrent-filemanager-media.git filemanager-media > ${log} 2>&1
                 ln -s /srv/rutorrent/plugins/filemanager-media/stream.php /srv/fancyindex/stream.php > /dev/null 2>&1
-                sed -i 's/$streampath .*/$streampath = $_SERVER['"'HTTP_HOST'"'] . '"'\/fancyindex\/stream.php'"';/g' /srv/rutorrent/plugins/filemanager-media/conf.php
+                sed -i 's/$streampath .*/$streampath = '"'https:\/\/'"'.$_SERVER['"'HTTP_HOST'"'] . '"'\/fancyindex\/stream.php'"';/g' /srv/rutorrent/plugins/filemanager-media/conf.php
                 ;;
             #fileupload)
             #svn co https://github.com/nelu/rutorrent-thirdparty-plugins/trunk/fileupload >/dev/null 2>&1
@@ -134,6 +134,9 @@ function _itplugs() {
             ratiocolor)
                 svn co https://github.com/Gyran/rutorrent-ratiocolor.git/trunk ratiocolor > /dev/null 2>&1
                 sed -i "s/changeWhat = \"cell-background\";/changeWhat = \"font\";/g" /srv/rutorrent/plugins/ratiocolor/init.js
+                ;;
+            *)
+                echo_error "$ is invalid"
                 ;;
         esac
     done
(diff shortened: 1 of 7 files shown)


stickz:
I would recommend pinning to this release instead. https://github.com/Novik/ruTorrent/releases/tag/v4.0-beta.1

Master has become a development branch on ruTorrent. Things are being committed to master which have regressions. This solution will resolve your problems and introduce a more stable environment for swizzin. This release has been heavily tested.

liaralabs:
I was thinking of pulling releases off the site, but that's a much better idea. ruTorrent was fairly stable at head for a while, but it's best to get things working at a stable point if regressions are going to be an issue. Thanks for the tip :smile:

scripts/rtx
```diff
@@ -103,11 +107,43 @@ function _itplugs() {
...
+                case $tag in
+                    3.*)
+                        svn co https://github.com/nelu/rutorrent-thirdparty-plugins/trunk/filemanager > /dev/null 2>&1
+                        chmod -R +x /srv/rutorrent/plugins/filemanager/scripts
+                        cat > /srv/rutorrent/plugins/filemanager/conf.php << FMCONF
+<?php
+\$fm['tempdir'] = '/tmp';
+\$fm['mkdperm'] = 755;
+// set with fullpath to binary or leave empty
+\$pathToExternals['rar'] = '$(which rar)';
```
stickz, line 119: It's better to use `command -v` here instead of `which`. This is a posix command (shell built-in) works on every Linux distribution with no dependencies required. It also has less overhead than `which` and is not depreciated.

scripts/rtx
```diff
@@ -103,51 +107,61 @@ function _itplugs() {
...
+                        svn co https://github.com/nelu/rutorrent-thirdparty-plugins/trunk/filemanager > /dev/null 2>&1
+                        chmod -R +x /srv/rutorrent/plugins/filemanager/scripts
+                        cat > /srv/rutorrent/plugins/filemanager/conf.php << FMCONF
 <?php
-
 \$fm['tempdir'] = '/tmp';
 \$fm['mkdperm'] = 755;
-
 // set with fullpath to binary or leave empty
 \$pathToExternals['rar'] = '$(which rar)';
```
stickz, line 119: It's better to use `command -v` here instead of `which`. This is a posix command (shell built-in) works on every Linux distribution with no dependencies required. It also has less overhead than `which` and is not deprecated.

liaralabs: We only support Debian/Ubuntu, so full posix compliance does not particularly concern me -- more understandable and readable code tends to be the language which I prefer and `which` is easier to understand for most than `command -v` (though there are a few `command -v`'s sprinkled in the project. As far as I'm aware, which was slated to be deprecated, but isn't going to be and the deprecation warning will be removed. 

https://lwn.net/Articles/874049/

Furthermore, I'm fairly certain the deprecation warnings only exist in Ubuntu short term releases right now, which we also don't support. If `which` does indeed get fully deprecated or moved out of debianutils then I will create a full PR which targets removing the use of `which` from the entire project.

stickz: okay sounds good. This is really only a hack for 3.x anyways. Eventually, everyone will be on 4.x and there will be no need for it anymore. The file pathing was fixed in 4.x with `command -v`. Install scripts won't need to populate these.

liaralabs: That's good to know -- I realized I didn't have to define the variables in the new FM install script. If they no longer need to be defined though then I'll go ahead and remove the `which` sections for the configuration of ruTorrent itself -- since install will only be supporting 4.0+ from this point

stickz (commented):
In the file `scripts/nginx/rutorrent.sh` on line 84 we have this:
`cachedEcho('{ "total": '.$total.', "free": '.$free.' }',"application/json");`
It needs to be changed to this for compatibility with ruTorrent v4.0:
`CachedEcho::send('{ "total": '.$total.', "free": '.$free.' }',"application/json");`

In the file `scripts/nginx/rutorrent.sh` on line 86 it starts like this:
`cachedEcho('{ "total": '.disk_total_space($topDirectory).', "free":` 
Likewise, it needs to start like this for compatibility with ruTorrent v4.0:
`CachedEcho::send('{ "total": '.disk_total_space($topDirectory).', "free": `

**Github repo this belongs to is also no longer valid:** https://github.com/QuickBox/quickbox_rutorrent-plugins

PR #831, opened 2022-02-13
fix(rutorrent): pin filemanager to last known working commit

liaralabs:
Filemanager is breaking ruTorrent.

Stop breaking ruTorrent by reverting commit of filemanager.

Issue filed upstream

### DIFF
--- a/scripts/rtx
+++ b/scripts/rtx
@@ -165,6 +165,7 @@ FMCONF
                         ;;
                     *)
                         git clone https://github.com/nelu/rutorrent-filemanager filemanager >> /dev/null 2>&1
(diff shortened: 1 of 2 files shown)


stickz:
@liaralabs I'm going to create a v4.0-beta2 pre-release of ruTorrent in the near future. These errors about the autoloader are converted to warnings. You also can set [$do_diagnostic = false](https://github.com/swizzin/swizzin/blob/develop/sources/functions/rutorrent#L82) to hush the warnings. I would just wait until you're done developing. Flipping this Boolean value will make things harder to debug and might not catch some things. Ideally, you'd probably want to remove the version pin of filemanager (to get the latest changes) when this happens.

PR #869, opened 2022-05-18
fix: arr nginx configs

brettpetch:
## Fixes issues: 
- Un-tracked issue: silences @bakerboy448 😸 

## Proposed Changes:
- Updates arr nginx configs to the most current "valid" configuration according to arr spec.

## Change Categories

- Bug fix


brettpetch:
@bakerboy448 can you please check this over boo?

bakerboy448 (approved):
Seems sane

drpoutine:
everything seems to check out.

small side note when using these
on the latest sonarr develop version `3.0.8.1520` using the `proxy_set_header Host $host;` was always returning an invalid host, had to change that to `proxy_set_header Host $proxy_host;` 

:thonk: after that everything ran like butter

liaralabs:
> on the latest sonarr develop version 3.0.8.1520 using the proxy_set_header Host $host; was always returning an invalid host, had to change that to proxy_set_header Host $proxy_host;

Yes this is correct -- this workaround is specific to mono. It can be dropped in any app that has migrated to dotnet core, but cannot be dropped on apps that still use mono (Sonarr).

liaralabs (approved):

drpoutine:
> 



> > on the latest sonarr develop version 3.0.8.1520 using the proxy_set_header Host $host; was always returning an invalid host, had to change that to proxy_set_header Host $proxy_host;
> 
> Yes this is correct -- this workaround is specific to mono. It can be dropped in any app that has migrated to dotnet core, but cannot be dropped on apps that still use mono (Sonarr).

I guess its unavoidable until sonarr switches off mono whenever that is. Til this these configs for nginx and their respective apps are correct and functional. :D

PR #875, opened 2022-05-29
Enhance rTorrent & ruTorrent with vmtouch

stickz:
### ruTorrent
This pull request loads and locks all the relevant ruTorrent files into memory using vmtouch. It improves the performance and stability of ruTorrent, especially during high disk usage scenarios.  **This feature requires only 15 megabytes of memory!**

**Examples of undesirable behavior fixed by this pull request:**
1. ruTorrent freezing during hash checking. The WebUI becomes unresponsive and unusable until completion.
2. ruTorrent taking longer than 20s to load or update the WebUI during high disk usage scenarios.

**Filters:**
1. It has a pattern filter to exclude all torrent files that ruTorrent saves. It is not necessary to load these into memory.
2. It has a file size filter to exclude all files greater than 125 kilobytes in size. The biggest file on ruTorrent is `jquery.flot.js` which is 121 kilobytes in size. I will get the minimum requirement down to 90 kilobytes (in the near future) by updating ruTorrent with a compressed version of this file. This size filter excludes large temporary files like the httprpc caches.

### rTorrent
This pull request loads and locks session files for rTorrent into memory using vmtouch. It makes the startup process for the software more efficient. Instead of reading the session files from the disk, they are read from memory instead.

**Filters:**
1. It has a pattern filter to only include the rTorrent and libTorrent session files. It skips unnecessary torrent files.
2. It two size filters. If there is more than 2GB of available memory, it will load session files up to 125KB. This only excludes session files from torrents which cannot be distributed.  Otherwise, it will optimize for low memory scenarios and only load session files up to 5KB. This will exclude extra some session files to save memory, increasing startup by up to a few seconds.

## Change Categories

- Bug fix 
- New feature 
- Change in `functions` 

## Checklist

- [X] Branch was made off the `develop` branch and the PR is targetting the `develop` branch
    - PR link: 
- [X] Changes to panel have been made OR are not necessary
    - PR link: 
- [X] Code conforms to project structure [(See more)](https://swizzin.ltd/dev/structure)
- [X] Testing was done
   - [X] Tests created or no new tests necessary
   - [X] Tests executed

### Architectures
| Jammy 		|	✅		|			|			|				|
| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|


stickz:
Requires ruTorrent master commit https://github.com/Novik/ruTorrent/commit/de014c5772f7d4d559967cb67148ee9babe08ebd or later. The maximum file size is reduced from 125KB to 90KB to reduce memory consumption. `jquery.flot.js` was compressed from 123KB to 36KB with this commit.

stickz:
> We’ll have to wait for a new beta release for it to be included — we only pull tags as it is now.
> 
> I’ll mark this as an external blocking dep until there’s been a tag on a release with that.
> 
> We stopped tracking master and moved to tracking tags because of the instability users were having with plugins.
> 
> Thanks for the work you’re doing :)
> 
> /b

Here's your new release: https://github.com/Novik/ruTorrent/releases/tag/v4.0-beta.2.1

We've tested the two major changes for over a month now. It's probably best to push everyone to this tag. There will be stable release of version 4.0 in the near future. The reason for the long delay was to give authors time to update their plugins.

stickz:
- Upgrade support has been added. It got a little bit messy with the sed commands and all the escaping, but it works.
- A bug has been resolved where it did not correctly replace 5K with 125K. A larger search and replace string is used now.

scripts/upgrade/rtorrent.sh
```diff
@@ -48,6 +72,9 @@ else
     rtorrent_apt
     echo_progress_done
 fi
+echo_progress_start "Updating systemd serivce "
+_systemd
```
static53, line 76: We'll need to call systemctl daemon-reload at some point to ensure the unit file changes are loaded.

scripts/install/rtorrent.sh
```diff
@@ -64,13 +64,22 @@ Type=forking
 KillMode=none
 User=%i
 ExecStartPre=-/bin/rm -f /home/%i/.sessions/rtorrent.lock
+ExecStartPre=-/usr/local/bin/vmtouch -i '*.torrent' -m 90K -dl /srv/rutorrent/
+ExecStartPre=-/usr/local/bin/vmtouch -I '*.torrent.libtorrent_resume' -I '*.torrent.rtorrent' -m 5K -dl /home/%i/.sessions/
```
static53, line 68: I noticed my vmtouch processes are piling up each time the service is restarted. Adding something like this is probably a good idea:

`ExecStopPost=pkill -f -u %i '/usr/local/bin/vmtouch.*/home/%i/.sessions'`

stickz:
I'm closing this pull request off. There are too many flaws with the implementation of this feature. It's not going to work in all environments. Even these issues are resolved, we run the risk of running out of file handles and breaking the operating system.

Issue #810, opened 2022-01-07
[BUG] 502 on jellyfin fresh install


riptide981:
### What happened?

Steps to reproduce:

1. Install package jellyfin
2. Go to https://<my_server_ip>/jellyfin
3. Error 502 Bad Gateway

Expected results:
Jellyfin first config web

### Swizzin commit

9ef0807

### What OS are you using?

Debian 11 (Bullseye)

### What architecture is your OS?

amd64

### Relevant logs and output

```plaintext
Swizzin is inside an LXC container.
root@swizzin-test:~# hostnamectl
   Static hostname: swizzin-test
         Icon name: computer-container
           Chassis: container
        Machine ID: 4e88bafb3c8941c2bd3a4d17201e6dcc
           Boot ID: e513983cddae4f47906e0217b2c62b1f
    Virtualization: lxc
  Operating System: Debian GNU/Linux 11 (bullseye)
            Kernel: Linux 4.19.0-18-amd64
      Architecture: x86-64

cat /var/log/nginx/error.log :
<date> <time> [error] 7871#7871: *2 connect() failed (111: Connection refused) while connecting to upstream, client: 192.168.1.84, server: _, request: "GET /jellyfin HTTP/2.0", upstream: "https://127.0.0.1:8920/jellyfin", host: "192.168.1.65", referrer: "https://192.168.1.65/"

I tried the Troubleshooting guide https://swizzin.ltd/guides/troubleshooting#-im-getting-a-502-when-accessing-the-server without success
```

brettpetch:
More information needed. Can you please collect relevant logs from `/root/logs/swizzin.log` and check out `systemctl status Jellyfin`

Additionally, can you confirm your Jellyfin port and verify it is the same in the reverse proxy?

Finally, ensure gpung is installed via APT prior to install of the application — this is a known issue that will be looked at soon.

riptide981:
relevant logs from /root/logs/swizzin.log (hope this help)
`root@swizzin-test:~# cat /root/logs/swizzin.log | grep jellyfin
jellyfin.sh
[Fri Jan 07 22:38:05] INFO      Installing jellyfin
Get:4 https://repo.jellyfin.org/debian bullseye InRelease [6,639 B]
Get:5 https://tor1.mirror.jellyfin.org/debian bullseye/main amd64 Packages [1,684 B]
  i965-va-driver intel-media-va-driver jellyfin-server jellyfin-web libaacs0
[...]

Issue #587, opened 2021-02-02
Jellyfin installation


LaMpiR:
## Bug description
SSL not installed and port for login use with nginx. Install through "box install jellyfin" as root.

## Reproduction steps
su
install box jellyfin

## Installed applications through swizzin
nginx, rutorrent, rtorrent, plex

## Expected behavior
To be accessible through panel on */jellyfin with SSL.

## Server Info
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.2 LTS
Release:        20.04
Codename:       focal

## Is Swizzin up to date?
Yes, box update command was run before that.

## Additional context
Swizzin uptodate. Wanted to try out the jellyfin.

Installed through "box install jellyfin" as a root. Installation was without any errors. Went to the panel, clicked o Jellyfin and got the error "Error processing request." from url */jellyfin.

Went to *:8920 but no ssl and it works. 

Is there anyway to install the ssl for the *:8920 and to change the url in the cotrol panel menu?

No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.2 LTS
Release:        20.04
Codename:       focal

LaMpiR:
Is this now added?

flying-sausages:
> Is there anyway to install the ssl for the *:8920 and to change the url in the cotrol panel menu?

Yes, read the docs.

You might want to look into the letsencrypt package in swizzin.

I don't really see a bug here, we're a lot better with giving support through discord if you want to have someone take a look at it with you

LaMpiR:
https://swizzin.ltd/applications/letsencrypt

I have read this and reinstalled it and reissued the certificate and it's still not applied to the jellyfin.
Jellyfin is using the Custom SSL certificate path:
/home/username/.ssl/username-self-signed.pfx
and not from /etc/nginx/ssl/hostname (there is not .pfx file here).

Using search https://swizzin.ltd/applications/letsencrypt doesn't work so I can't find anything with SSL additionaly.

flying-sausages:
Letsencrypt will only affect the `/jellyfin` endpoint, not the application's port. I'm still unsure what exactly you're trying to access this way and why.

Did you see this? https://swizzin.ltd/applications/jellyfin#unable-to-connect-using-apps

LaMpiR:
> Letsencrypt will only affect the `/jellyfin` endpoint, not the application's port. I'm still unsure what exactly you're trying to access this way and why.
> 
[...]

PR #894, opened 2022-07-31
Let `nginx` handle SSL and redirect to HTTP

the-mann:
## Description
Out of the box Jellyfin won't connect unless you connect directly (ie `http://<ip-address>:8096`). `http://<ip-address>/jellyfin` will give a `502` because HTTPS isn't set up by default on Jellyfin (nor should it be Jellyfin's responsibility to handle SSL/HTTPS, that's the job of the reverse proxy). Therefore by referring to the HTTP port of Jellyfin, we can avoid the chicken egg scenario and let `nginx` do what it's designed to do.

## Fixes issues: 
- Issue #810 
- Related to #587 

## Proposed Changes:
- Changed Nginx to proxy to HTTP Jellyfin rather than HTTPS Jellyfin, as we already have HTTPS from Client to Nginx, which makes HTTPS between reverse proxy and Jellyfin redundant.

## Change Categories
- Bug fix 
- I doubt it would be a breaking change, but I suppose it could be if some users have modified their Jellyfin config to only support HTTPS rather than letting `nginx` handling SSL 
- 
## Checklist

    - PR link: 
    - PR link: 
- [x] Testing was done
   - [ ] Tests executed

### DIFF
--- a/scripts/nginx/jellyfin.sh
+++ b/scripts/nginx/jellyfin.sh
@@ -24,7 +24,7 @@ fi
 # Create our nginx application conf for jellyfin
 cat > /etc/nginx/apps/jellyfin.conf <<- NGINXCONF
 	location /jellyfin {
-		proxy_pass https://127.0.0.1:8920;
+		proxy_pass http://127.0.0.1:8096;
 		#
 		proxy_pass_request_headers on;
 		#


liaralabs:
The code was contributed this way and the method of using ssl in the proxy was discussed at the time. I let it be since ssl certificates are created for jellyfin to use during the spin up of the application and this technically doesn't actually create issues. There aren't any bugs or major issues with communicating via SSL over localhost and everything is present in order to do so.

The current issues related to 502 errors were a result of the network.xml file not being created consistently which has a cascading effect of not allowing jellyfin to be configured correctly for the proxy configuration.

If there's actually a bug I'd like to see proof/evidence, rather than just linking previous issues which have already been closed with other fixes.

brettpetch:
> The code was contributed this way and the method of using ssl in the proxy was discussed at the time. I let it be since ssl certificates are created for jellyfin to use during the spin up of the application and this technically doesn't actually create issues. There aren't any bugs or major issues with communicating via SSL over localhost and everything is present in order to do so.
> 
> The current issues related to 502 errors were a result of the network.xml file not being created consistently which has a cascading effect of not allowing jellyfin to be configured correctly for the proxy configuration.
> 
> If there's actually a bug I'd like to see proof/evidence, rather than just linking previous issues which have already been closed with other fixes.

As a quick followup on this, a bug has been found in the Nginx script that fires before network.xml is generated on Debian systems. I’ll be issuing a patch to fix this later in the day/week.

rtrive:
I think the same problem happens with Sickchill. When you try to reach it, it returns 404 meanwhile accessing through tunnel show the dashboard

PR #941, opened 2022-12-28
Fix header location for nextcloud installation

herrwusel:
Currently the headers are in the wrong place in the nginx config.

## Description
In the current status the recommended headers for a nextcloud installation are not set (can be checked with `curl -I tld/nextcloud`). This commit fixes the headers.

## Fixes issues: 
- Un-tracked issue

## Proposed Changes:
- Put the headers in the right position in the nginx config.

## Change Categories

- Bug fix 

## Checklist

- [ ] Branch was made off the `develop` branch and the PR is targetting the `develop` branch
    - PR link: 
    - PR link: 
- [ ] Testing was done
   - [ ] Tests executed

### DIFF
--- a/scripts/install/nextcloud.sh
+++ b/scripts/install/nextcloud.sh
@@ -132,6 +132,25 @@ location = /.well-known/caldav {
 location /.well-known/acme-challenge { }
 
 location ^~ /nextcloud {
+    add_header Cache-Control "public, max-age=15778463";
+    # Add headers to serve security related headers  (It is intended
+    # to have those duplicated to the ones above)
+    # Before enabling Strict-Transport-Security headers please read
+    # into this topic first.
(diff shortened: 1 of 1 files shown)


liaralabs:
Hi @herrwusel -- I would be happy to include this. However, in the current state, this will not update current installations. Can you please add an updater script that will check if existing installations need an edit and to apply the edit if found to be out of date?

herrwusel:
Oh, of course. I will have a look. Are there any previous updater scripts you would recommend?

liaralabs:
Apologies on the very late response here.

> Is it enough if I change the headers' location there? If not, I'm uncertain if I understand correctly which update functionality you are speaking off.

Yes it would be sufficient to check if the nextcloud update file contains the location block that shouldn't exist and use that as a trigger for the new code

PR #962, opened 2023-03-01
rTorrent: Installation speed improvements

stickz:
Compile libtorrent and rtorrent with level 1 optimizations if there are 2 or less processor threads available.
- This drastically reduces build times on systems such as virtual machines. It is still better than level 0 repo.

Compile xmlrpc, libtorrent and rtorrent with piping if at least 512 MB of memory is available.
- This greatly reduces build times by using memory instead of temporary files for compilation.

Compile xmlrpc, libtorrent and rtorrent without warning messages.
- Warning messages do not impact functionality of the software. 
- Disabling these reduces compile times by eliminating overhead.

Retrieve libtorrent and rtorrent faster using curl instead of wget.
- This is already installed because it's required for rtorrent.

Don't verbose messages about libtorrent and rtorrent extraction.
- The extraction process if faster if we don't list the files.
- The log file is also cleaner and will still show errors.

### DIFF
--- a/scripts/install/rtorrent.sh
+++ b/scripts/install/rtorrent.sh
@@ -98,6 +98,7 @@ if [[ -n $noexec ]]; then
 fi
 depends_rtorrent
 if [[ ! $rtorrentver == repo ]]; then
+    configure_rtorrent
     echo_progress_start "Building xmlrpc-c from source"
     build_xmlrpc-c
     echo_progress_done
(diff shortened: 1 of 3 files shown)


liaralabs:
Thanks for all the effort recently. This is almost looking good to me, but I think I would prefer if we only repeat the code once -- this way we can maintain the build arguments in a single place. I think a pre-build step might fit nicely here, then we just need to ensure the upgrade script includes the new function as well.

stickz:
@liaralabs The requested change has been implemented and tested. `configure_rtorrent` will configure the build variables for all newly added rtorrent items. It took 90s (30s per item) to compile on an AMD four core virtual machine with `-pipe -flto -O2`.

liaralabs:
It shouldn't cause issues but exporting the variables in the configure function shouldn't actually required since all these functions are running in the same bash shell. I'm not going to request changes but just a note.

PR #967, opened 2023-03-24
rTorrent: Add install advisory & sink repo release

stickz:
This pull request adds an install advisory for rTorrent and moves the repo release to the bottom of the list.

### DIFF
--- a/sources/functions/rtorrent
+++ b/sources/functions/rtorrent
@@ -9,8 +9,8 @@ function whiptail_rtorrent() {
         repov=$(get_candidate_version rtorrent)
 
         function=$(whiptail --title "Choose an rTorrent version" --menu "All versions other than repo will be locally compiled from source" --ok-button "Continue" 14 50 5 \
+            0.9.8 "(Recommended)" \
             Repo "(${repov})" \
-            0.9.8 "" \
             0.9.7 "" \
             0.9.6 "" 3>&1 1>&2 2>&3) || {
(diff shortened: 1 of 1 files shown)


liaralabs:
I'm hesitant to push this through for the following reasons:

1. We don't recommend versions of other clients, the user is generally left to choose what's appropriate for themselves
2. It's already at the top of the list (aside from repo)
3. I disagree with 0.9.8 being the best version available for all use-cases

I would sooner rather make the changes if some updating here is requested:

- Change the whiptail box wording to something like "We recommend compiling over the distro packages"
- Move repo to the bottom of the list

stickz:
@liaralabs What do you think about these changes? I added a new whiptail message box to advise users about compiling locally from source. And sunk the repo release to the bottom of the list. It no longer recommends a particular version as requested.

- I don't know why the `sh-checker` is failing. I tested these changes and they worked fine.
- The reason for the message box is to give users additional information that won't fit in the selection menu.

liaralabs:
sh-checker is validating that the code meets the formatting standards -- it's less to do about syntax (though it should catch those as well). I use vs-code with the shellformat extension. The projects settings will automatically format your files correctly on save.

Otherwise, you might need to pass your files through shellformat manually with the following flags.

```
-ci -sr -i 4 -ln bash
```

stickz:
> Otherwise, you might need to pass your files through shellformat manually with the following flags.

This is done. The file adheres to formatting standards now.

PR #973, opened 2023-04-04
rTorrent: Resolve lockfile crash

stickz:
Backports fix for lock file crash merged into rTorrent master. https://github.com/rakshasa/rtorrent/pull/1169


stickz:
Tested on Ubuntu 22.04 LTS with rTorrent 0.9.8 source build. Confirmed to be working. File was not changed in 12 years on rTorrent. It should be fully supported and working on 0.9.7 and 0.9.6 as well.

PR #975, opened 2023-04-04
rTorrent: Install faster with ramdisk

stickz:
This commit mounts a 1 GB tmpfs ram disk for rtorrent installations.  Four gigabytes of system memory is required. Increases installation speed and efficiency.

It was required to refactor the build paths. The ram disk needs to be mounted in a non conflicting location.

Also, it creates a new `rtorrent.log` file for the rtorrent installation. The log file is written to the ram disk, then copied to the logs folder afterwards. This resolves the megabyte log file problem with multiple installs and reduces disk writes.

Changes were tested on an 8 core Ubuntu 22.04 LTS VPS with 16GB of RAM. rTorrent source 0.9.8 was built. The entire installation process of rTorrent (from start to finish) with `box install rtorrent` took 82 seconds. Faster install times are anticipated with dedicated resources. There was significant shell overhead with the virtualized environment slowing things down.

PR #977, opened 2023-04-10
Fix: xmlrpc-c will not compile on newer systems

zottelbeyer:
## Description

XML-RPC-C fails to configure on my QNAP NAS and updating the config.guess / config.sub fixes it.

## Fixes issues: 
- Un-tracked issue...

## Proposed Changes:
- download updated config.guess and config.sub before configuring xml-rpc-c

## Change Categories

- Bug fix 

## Checklist

- [ ] Testing was done
   - [ ] Tests executed

### Architectures

| Jammy 		|			|			|			|				|
| Focal 		|			|			|			|				|
| Bionic		|			|			|			|				|
| Bullseye		|			|			|	✅		|				|
| Buster		|			|			|			|				|
| Stretch		|			|			|			|				|

### DIFF
--- a/sources/functions/rtorrent
+++ b/sources/functions/rtorrent
@@ -104,6 +104,8 @@ function build_xmlrpc-c() {
     XMLRPC_REV=2954
     svn co http://svn.code.sf.net/p/xmlrpc-c/code/advanced@$XMLRPC_REV /tmp/xmlrpc-c >> $log 2>&1 || { svn co https://github.com/mirror/xmlrpc-c/trunk/advanced@$XMLRPC_REV /tmp/xmlrpc-c >> $log 2>&1; }
     cd /tmp/xmlrpc-c
+    curl -sL "http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD" -o ./config.guess >> $log 2>&1
+    curl -sL "http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD" -o ./config.sub >> $log 2>&1
     ./configure --prefix=/usr --disable-cplusplus --disable-wininet-client --disable-libwww-client >> $log 2>&1 || {
         echo_error "Something went wrong while configuring xmlrpc"
         exit 1


stickz:
Could you try retrieving xmlrpc-c from GitHub? And compiling from the `stable` folder?
https://github.com/mirror/xmlrpc-c

This is not a good solution because it adds overhead. We should retrieve it from GitHub using `git clone` instead. It's faster and more reliable. Furthermore, if this link becomes invalid (which often happens on the GNU website) the install script will fail.

stickz:
Alternatively, if the developer version of xmlrpc-c is broken, we could distribute these two files with the project as patches. Then copy them to the xmlrpc-c installation folder. It's more important to keep things working, then have the latest version.

zottelbeyer:
> Could you try retrieving xmlrpc-c from GitHub? And compiling from the `stable` folder? https://github.com/mirror/xmlrpc-c
> 
> This is not a good solution because it adds overhead. We should retrieve it from GitHub using `git clone` instead. It's faster and more reliable. Furthermore, if this link becomes invalid (which often happens on the GNU website) the install script will fail.

the stable version from git compiles fine for me.

> Alternatively, if the developer version of xmlrpc-c is broken, we could distribute these two files with the project as patches. 
> Then copy them to the xmlrpc-c installation folder. It's more important to keep things working, then have the latest version.

Yes, I see the point in that. The affected hardware is a TS-832PXU which I believe was released in 2021.

It would probably be enough to just update the revision to a later version?

stickz:
> Yes, I see the point in that. The affected hardware is a TS-832PXU which I believe was released in 2021.
> 
> It would probably be enough to just update the revision to a later version?

I would ask @liaralabs for thoughts on using the GitHub mirror. We can reset the version to pin the latest commit. `svn co` has performance issues. I was thinking about changing it last release cycle, but left it because there wasn't a good reason.

stickz:
Even if `git clone` and `git reset --hard` takes a few hundred milliseconds longer, it uses a fraction of the CPU resources. This has a significant advantage when deploying to weaker systems like Raspberry PIs. Furthermore, I have plans to clone source files and download plus extract tar files asynchronously. It will save seconds and work everywhere with `git`.

liaralabs:
The version of XMLRPC-C was carefully chosen because being a bit more careless with xmlrpc-c versioning has led to weird quirks and bugs which have manifested in the form of rtorrent crashes. I don't know if rtorrent or xmlrpc are to blame for the crashes; however, rakshasa hasn't released any stable versions of rTorrent since 2019 (our xmlrpc version is from 2017) so I'm still a bit skeptical that newer versions won't just regress and bring these quirks back into the fold.

The only way to find out though is to test.

I don't really have a preference for git vs svn -- svn is the official way of accessing the source code though and the authors certainly don't promote or encourage users to use the github mirror in their documentation.

I would only say that I'm generally against tracking the latest commit and would prefer if requests to change the version of XMLRPC-C are tested before blindly choosing a new version. Tracking head used to be the previous behavior before a version pin was put into place and I don't intend to revert this behavior.

Case and point: debian and ubuntu have both been using 1.33.14 since as far back as Bionic/Stretch and newer OSes don't utilize newer versions. I'm not really in a hurry to bump the version of this dependency, to be honest.

stickz:
I'm waiting on #975 merge to patch this problem. It's not compatible because the installation directory has changed.

stickz:
Never mind actually. I'm going to apply the patch to the current directory. It can be changed afterwards for async tasks.

PR #981, opened 2023-04-17
rTorrent: Patch three memory leaks

stickz:
This commit patches three memory leaks with the rTorrent client to improve memory stability. These three patches were selected because of their known stability. I these changes on ARM64 by downloading a torrent and compiling from source. The compile log succeeded for all source versions (0.9.6, 0.9.7, 0.9.8) and the torrent was successfully downloaded. There are no known issues reported by other users with these changes. I've also tested them personally for an entire year to ensure they are stable.

**rTorrent version 0.9.8**
- Patches a memory leak with the text console on a rTorrent screen.

**rTorrent all versions**
- Patches two memory leaks related to how rTorrent handles commands in `rtorrent.rc`.

The line numbers have changes on different source versions, which is why multiple patch files are required. The patches are created in a new folder to make it easier to find other patches and to avoid confusion. Each different type of patch is separated.

Memory stability is a big problem for rTorrent. Reducing memory leaks allows the torrent client to run longer without a restart.

### DIFF
--- a/sources/functions/rtorrent
+++ b/sources/functions/rtorrent
@@ -202,11 +202,24 @@ function build_rtorrent() {
     patch -p1 < /etc/swizzin/sources/patches/rtorrent/lockfile-fix.patch >> "$log" 2>&1
     #apply xmlrpc-fix to all rtorrents
     patch -p1 < /etc/swizzin/sources/patches/rtorrent/xmlrpc-fix.patch >> "$log" 2>&1
-    #use pkgconfig for cppunit if 0.9.6
     stdc=
-    if [[ ${rtorrentver} == "0.9.6" ]]; then
+    if [[ ${rtorrentver} == "0.9.8" ]]; then
+        # Patch download list memory leak with rTorrent screen specific with version 0.9.8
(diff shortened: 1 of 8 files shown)


stickz:
I marked this pull request as a draft PR. It's required to perform work to make it compatible with the UDNS pull request.

sources/functions/rtorrent
```diff
@@ -202,11 +202,24 @@ function build_rtorrent() {
     patch -p1 < /etc/swizzin/sources/patches/rtorrent/lockfile-fix.patch >> "$log" 2>&1
     #apply xmlrpc-fix to all rtorrents
     patch -p1 < /etc/swizzin/sources/patches/rtorrent/xmlrpc-fix.patch >> "$log" 2>&1
-    #use pkgconfig for cppunit if 0.9.6
     stdc=
-    if [[ ${rtorrentver} == "0.9.6" ]]; then
+    if [[ ${rtorrentver} == "0.9.8" ]]; then
```
liaralabs, line 206: This would be the perfect place for a case statement -- things are getting a bit more verbose and the case would help clean up readability.

stickz:
Closing this off. Replaced by #1084.

PR #982, opened 2023-04-17
rTorrent: Add UDNS install option

stickz:
This commit adds an option to install a stability patch for UDP trackers on rTorrent. It's back ported against version 0.9.8.

With this patch, the user will be able to distribute thousands of torrents with UDP trackers, without running into software stability issues. That being said, there are still two unsolved race conditions. This patch is not at fault for causing the crashes.

Install advisory is updated to inform users about the option. libudns is built from source for compatibility with Debian. It takes a few seconds and is only built if the option is enabled.

Changes are tested on Ubuntu 22.04LTS with ARM64. They are confirmed to be working correctly by testing and build logs.

### DIFF
--- a/sources/functions/rtorrent
+++ b/sources/functions/rtorrent
@@ -8,12 +8,13 @@ function whiptail_rtorrent() {
     if [[ -z $rtorrentver ]] && [[ -z $1 ]]; then
         repov=$(get_candidate_version rtorrent)
 
-        whiptail --title "rTorrent Install Advisory" --msgbox "We recommend rTorrent version selection instead of repo (distro) releases. They will compile additional performance and stability improvements in 90s." 15 50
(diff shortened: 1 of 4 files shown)


stickz:
**rTorrent: Fix libtorrent scanf crash** has passed the build process. This patch resolves a rTorrent crash related to **scanf** being called tens of thousands of times during startup. It caches the results into a vector and calls it once per tracker instead.

stickz:
**rTorrent: Disable forced tracker startup scrape** has passed the build procress. This patch resolves a rTorrent crash related to thousands of tracker updates being performed on startup. The thread times out after five minutes and fails to start the software.

The solution to this problem is to not perform tracker updates on startup. Instead, they are queued after startup using the normal method of tracker updates intervals. It may take slightly longer for tracker updates, however startup times are now instantaneous.

PR #983, opened 2023-04-18
AutoDL: Fix wget missing URL error

stickz:
The `--quiet` parameter cannot be used. We must use `-q` before `-O` instead, to achieve the intended no verbose effect.

```...     Downloading autodl source code
wget: missing URL
Usage: wget [OPTION]... [URL]...```

### DIFF
--- a/scripts/install/autodl.sh
+++ b/scripts/install/autodl.sh
@@ -27,7 +27,7 @@ function _installautodl() {
 
 function _autoconf() {
     echo_progress_start "Downloading autodl source code"
-    curl -sL http://git.io/vlcND | grep -Po '(?<="browser_download_url":).*?[^\\].zip"' | sed 's/"//g' | xargs wget -O /tmp/autodl-irssi.zip --quiet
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by liaralabs** [comment]
This is not an actual problem

```
Logging and input file:
  -o,  --output-file=FILE          log messages to FILE
  -a,  --append-output=FILE        append messages to FILE
  -d,  --debug                     print lots of debugging information
  -q,  --quiet                     quiet (no output)
```

I just tested bullseye and jammy -- these both work currently.

I would suggest we modernize this by using the github latest version function and then use a curl / error checking along the way

stickz:
I tested this on Jammy and it failed. When I ran this command in terminal, I was able to re-simulate the same error Pichu experienced. You can't specify `--quiet` after the file path. Arguments for `wget` need to be specified beforehand.

`curl -sL http://git.io/vlcND | grep -Po '(?<="browser_download_url":).*?[^\\].zip"' | sed 's/"//g' | xargs wget -O /tmp/autodl-irssi.zip --quiet`

liaralabs:
If that's the case then why am I running successful tests under jammy as-is?

stickz:
> If that's the case then why am I running successful tests under jammy as-is?

That's a good question. I'm not longer able to re-simulate the bug. Let's close this off and rewrite the script instead.

PR #984, opened 2023-04-18
ruTorrent: Set $localHostedMode to true

stickz:
It was introduced and set to false when testing ruTorrent version 4.0.1. The regressions were since resolved in ruTorrent version 4.0.2. See change-log here: https://github.com/Novik/ruTorrent/releases/tag/v4.0.2

The feature purpose is to improve loading times when ruTorrent is hosted on the SAME machine as rTorrent. It performs plugin dependency checks through shell and caches the results. This is a more viable option than executing XMLRPC commands, which is designed be performed for remote hosts. XMLRPC is bound by various bottlenecks such as disk i/o resources.

### DIFF
--- a/sources/functions/rutorrent
+++ b/sources/functions/rutorrent
@@ -113,7 +113,7 @@ function rutorrent_install() {
 		"stat"	=> '',			// Something like /usr/bin/stat. If empty, will be found in PATH.
 	);
 	
-	$localHostedMode = false; 		// Set to true if rTorrent is hosted on the SAME machine as ruTorrent
+	$localHostedMode = true; 		// Set to true if rTorrent is hosted on the SAME machine as ruTorrent
 	
 	$cachedPluginLoading = false;	// Set to true to enable rapid cached loading of ruTorrent plugins	
 	$pluginJSCacheExpire = 3*60;	// Sets duration ruTorrent plugin javascript cache is valid for in minutes


sources/functions/rutorrent
```diff
@@ -113,7 +113,7 @@ function rutorrent_install() {
 		"stat"	=> '',			// Something like /usr/bin/stat. If empty, will be found in PATH.
 	);
 	
-	$localHostedMode = false; 		// Set to true if rTorrent is hosted on the SAME machine as ruTorrent
+	$localHostedMode = true; 		// Set to true if rTorrent is hosted on the SAME machine as ruTorrent
```
flying-sausages, line 116: I think it would be nice to apply this to previous installations as well, any chance you could write an update function for this, please?

stickz: `ruTorrent` does not support updating currently. The user is required to remove and reinstall. In the future, when updating support is added, these changes will apply by default.

flying-sausages: I'm referring to our own update files, namely https://github.com/swizzin/swizzin/blob/master/scripts/update/rutorrent.sh

This is where we put any config changes that need to be applied to old installs.

stickz: This change is optional. It doesn't need to be applied to old installs. It will use the `XMLRPC` method if set to `false`. I don't think we should write an update function for this. It's more code to maintain, that will eventually need depreciated.

flying-sausages (approved):

PR #986, opened 2023-04-23
AutoDL: Use consistent install/upgrade methods

stickz:
We should use consistent install and upgrade methods for autodl. This implementation is cleaner than the previous one and prevents possible errors from occurring. https://github.com/swizzin/swizzin/blob/master/scripts/upgrade/autodl.sh#L8-L11


flying-sausages (approved):
LGTM

PR #987, opened 2023-04-23
Setup: Add BBR congestion control configuration

stickz:
This commit adds a feature to enable the BBR congestion control algorithm. This increases throughput of TCP connections thanks to clever engineering by Google.


brettpetch:
This will break unattended installation. Needs to have a switch that can be specified in the env file at minimum. 

Some vendors that use our script may use custom kernels, so you may want to check the current congestion control algorithm first, then list the available congestion control algorithms via `sysctl` instead, enabling only if both those requirements are satisfied.

stickz:
@brettpetch This pull request is ready for review again. I tested it on Ubuntu 22.04 LTS on ARM.
- Added `--enablebbr` command line option and updated readme.
- Added `BBR_CONGESTION_CONTROL` variable for env file.

The solution I used for unknowledgeable users is to check if they are running a generic Linux Kernel. An advanced user can override this feature by enabling it through the command line or env file. The kernel module is not loaded until we enable it, so we can't easily check if the feature is supported. But we do know that it's enabled by default and all generic kernel support it.

Unattended setups or custom kernels must be enabled manually. It will only prompt the user if the Linux Kernel is generic.

setup.sh
```diff
@@ -221,6 +225,37 @@ function _preparation() {
...
+function _checkbbr() {
+    # If bbr is enabled through unattended setup mark it as enabled
+    if [[ -n $BBR_CONGESTION_CONTROL ]]; then
+        [[ $BBR_CONGESTION_CONTROL = "yes" ]] && enablebbr=true
+    fi
+    local ready=false
+    # If enablebbr was set or the kernel is generic
+    if [[ $enablebbr = "true" ]] || [[ $(uname -r | grep -c generic) -ge 1 ]]; then
+        # If the sysctl file does not already contain a congestion control algorithm
+        if [[ $(grep -c net.ipv4.tcp_congestion_control /etc/sysctl.conf) -le 0 ]]; then
```
brettpetch, line 237:
```bash
# todo: log old
old_cc=$(sysctl net.ipv4.tcp_congestion_control | awk -F= '{print $2}' | sed -e s/\^\\s//)
available_cc=$(sysctl net.ipv4.tcp_available_congestion_control)

# todo: log available
if [[ $available_cc != *"bbr"* ]]; then
    modprobe tcp_bbr
    available_cc=$(sysctl net.ipv4.tcp_available_congestion_control)
    if [[ $available_cc != *"bbr"* ]]; then
        echo_info "Looks like your kernel doesn't support BBR. This change will no proceed."
    fi
fi

if [[ $available_cc == *"bbr"* ]]; then
  echo "net.core.default_qdisc=fq" | tee -a /etc/sysctl.d/10-custom-kernel-bbr.conf
  echo "net.ipv4.tcp_congestion_control=bbr" | tee -a /etc/sysctl.d/10-custom-kernel-bbr.conf
  sysctl -p || {
    echo_error "Failed to apply BBR congestion control, rolling back"
    rm_if_exists /etc/sysctl.d/10-custom-kernel-bbr.conf
  }
fi
```

brettpetch (changes requested):
needs a check to ensure that bbr is in the kernel.

stickz:
Alright, I improved the script to check if the Linux Kernel supports `BBR` and `FQ`. It should be pretty robust now.

stickz:
Feature is tested on Ubuntu 22.04LTS ARM64. It's perfected now to use `55-swizzin.conf` to override `50-default.conf`. This is required to ensure BBR Congestion Control has the proper Fair Queue Scheduler successfully enabled to avoid packet loss.

PR #989, opened 2023-04-25
ruTorrent: Install ffmpeg by default

stickz:
ffmpeg is required for the screenshots plugin. Many users want to use this plugin, so it's not a good idea to gut the functionality. It's supported on all platforms and is a tiny binary file.

This commit makes the ffmpeg install script modular, so it's easier to maintain and changes apply to ruTorrent. It only installs ffmpeg if it's not already installed.

These changes are tested on Ubuntu 22.04 LTS ARM64. They are confirmed to be working and the screenshots plugin is now functional out of the box. When users deploy ruTorrent, everything will work as intended!

### DIFF
--- a/sources/functions/ffmpeg
+++ b/sources/functions/ffmpeg
@@ -12,5 +12,12 @@
 #   including (via compiler) GPL-licensed code must also be made available
 #   under the GPL along with build & install instructions.
 
-. /etc/swizzin/sources/functions/ffmpeg
-ffmpeg_install
+ffmpeg_install() {
+    export distribution=$(lsb_release -is)
+    export release=$(lsb_release -rs)
+    export codename=$(lsb_release -cs)
+
+    apt_install ffmpeg
+
+    touch /install/.ffmpeg.lock
+}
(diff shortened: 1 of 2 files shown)


sources/functions/ffmpeg
```diff
@@ -0,0 +1,23 @@
...
+#
+#   You may copy, distribute and modify the software as long as you track
+#   changes/dates in source files. Any modifications to our software
+#   including (via compiler) GPL-licensed code must also be made available
+#   under the GPL along with build & install instructions.
+
+ffmpeg_install() {
+    export distribution=$(lsb_release -is)
+    export release=$(lsb_release -rs)
+    export codename=$(lsb_release -cs)
```
flying-sausages, line 18: @liaralabs can these go?

stickz: Those are taken from the original install script. It's just simple refactoring. I disagree with making changes not related to the pull request. It's harder to track down regressions. We should remove with a new commit if required.

flying-sausages: Sure, why not

flying-sausages:
@liaralabs any opposition to not install ffmpeg with rutorrent?

PR #990, opened 2023-04-25
ruTorrent: ARM64 support for file manager

stickz:
This commit drops rar archive creation on ARM64 because it's not supported. It only makes the adjustment to ARM to keep the x86 platform fully functional. This small change makes the file manager plugin work properly, without errors.

Changes tested on Ubuntu 22.04 LTS ARM64 and confirmed to be working. This makes ruTorrent function out of the box on ARM.

### DIFF
--- a/sources/functions/rutorrent
+++ b/sources/functions/rutorrent
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 function rutorrent_install() {
     apt_install sox geoip-database p7zip-full zip unzip
 
@@ -35,6 +37,7 @@ function rutorrent_install() {
 
     if [[ ! -d /srv/rutorrent/plugins/filemanager ]]; then
         git clone https://github.com/nelu/rutorrent-filemanager /srv/rutorrent/plugins/filemanager >> ${log} 2>&1 || { echo_error "git of autodl plugin to main plugins seems to have failed"; }
+        rutorrent_fm_config
         #pinning filemanager should no longer be required
         #git -C /srv/rutorrent/plugins/filemanager checkout 234f5f20841ad3d1e3c095e6c6954a875fc8a6ea >> ${log} 2>&1
     fi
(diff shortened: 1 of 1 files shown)


sources/functions/rutorrent
flying-sausages, line 1: Could you throw a shebang at the top while you're at it please?

sources/functions/rutorrent
```diff
@@ -146,6 +147,53 @@ function rutorrent_install() {
 RUC
 }
 
+function rutorrent_fm_config() {
```
flying-sausages, line 150: Would it be simpler to just `sed` the required changes into the file after the installation? I'd rather make changes post-install so that we won't have to keep updating this hardcoded config in mind.

stickz: The configuration we are editing would look terrible with `sed`. The user would not be able to understand how to edit.
```
    $config['archive']['type'] = [
        '7z' => [
            'bin' => '7zip',
            'compression' => [1, 5, 9],
        ]];
```

flying-sausages: Ah, when I say "after installation", I mean within the scope of the script, and not something that a user needs to do manually. So you'd let the "standard" config install itself first, and once you know that file is present, you just sed it within the install script.

stickz: I'm saying the user may want to edit this file. We should keep the formatting intact for them. `sed` is not designed to replace things like this. It works better for one liners not arrays.

jvacek: Yeah they can edit it as much as they can post-install, nothing wrong with that, we can still make changes to it either way.

We could also just append that config at the end of the file, there's ways

flying-sausages: Yeah they can edit it as much as they can post-install, nothing wrong with that, we can still make changes to it either way.

We could also just append that config at the end of the file, there's ways

stickz: Okay, I will echo text to the end of the file next week. It's a config variable that can be reassigned.

stickz: @flying-sausages @liaralabs We can't use `sed` or `echo` here. It does not work. It's required to use the `cat` method like all the other files. If this is too messy in bash, we can put it into the `patches` folder and replace the file. 

Please advise about next steps. I just wasted an hour trying to fix something that wasn't broken.

liaralabs: I'm still a bit hesitant to merge this as-is. `cat`ing entire config files isn't my favourite approach to keeping things easy to maintain long term.

I'm trying to understand the difference between this config and the one shipped for amd64. Is it basically the removal of these lines?

```
$pathToExternals['rar'] = '';
and
    'rar' => [
        'bin' => 'rar',
        'compression' => [0, 3, 5],
    ]
```

stickz: Yes, that is correct. We are removing lines for the unsupported `rar` binary.

PR #991, opened 2023-04-26
rTorrent: Compile curl with c-ares

stickz:
This pull request enhances curl for rTorrent by recompiling it with c-ares support for threaded DNS lookups. The latest stable release of curl from Debian Bullseye back-ports is used. The latest upstream release of c-ares is used.

Tested on Ubuntu 22.04LTS ARM64 and confirmed to be working.

### DIFF
--- a/scripts/install/rtorrent.sh
+++ b/scripts/install/rtorrent.sh
@@ -77,6 +77,7 @@ EOF
 export DEBIAN_FRONTEND=noninteractive
 
 . /etc/swizzin/sources/functions/rtorrent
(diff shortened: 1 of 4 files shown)


liaralabs:
I'm not going to block merging this improvement, but as the compile of rTorrent is becoming more complex, I would like to see improvements made in the area of avoiding recompiling packages during the upgrades which don't need to be touched. I.e if the version of curl hasn't changed, we don't need to run it through the compile pipeline again.

xmlrpc additionally pretty much never changes (I know this is technical debt that isn't related to this contribution)

PR #992, opened 2023-04-26
rTorrent: Upgrade GCC if supported

stickz:
This script upgrades GCC on Ubuntu 20.04 or 22.04 LTS using packages. It improves the performance of software compiled with GCC. This task is performed relatively quickly. 
- Ubuntu 20.04 LTS seamlessly goes from GCC 9 to 10. Ubuntu 22.04 LTS seamlessly goes from GCC 11 to 12. 
- update-alternatives are configured to allow smooth transition from Ubuntu 20.04 to 22.04 LTS. The user just needs to run the script again for the newer version to take priority over the old. If they would like to recompile software.
- Manual upgrade option is added. This task is performed automatically if rTorrent is being compiled from source. 
- Debian has chosen not to provide packages for upgrading GCC. Therefore, it's not possible to seamlessly support this OS. It would take hours to manually compile GCC.

This has been tested on Ubuntu 22.04 LTS ARM64 and confirmed to be working.

### DIFF
--- a/sources/functions/gcc
+++ b/sources/functions/gcc
@@ -1,0 +1,49 @@
+#!/bin/bash
+# Functions to upgrade GCC on Ubuntu
+# Stickz for Swizzin 2023
+# GPLv3 Applies
+
+Upgrade_GCC_Ubuntu() {
+    echo_info "Checking OS for GCC upgrade"
+    if [[ "$(_os_codename)" = "focal" ]]; then
+        Upgrade_GCC10_Focal
+    elif [[ "$(_os_codename)" = "jammy" ]]; then
+        Upgrade_GCC12_Jammy
+    else
+        echo_info "Skipping GCC upgrade due to unsupported OS"
+    fi
+}
+
+Upgrade_GCC10_Focal() {
+    local gccten=$(gcc --version | grep -c "Ubuntu 10")
+    if [[ $gccten -le 0 ]]; then
+        echo_info "Ubuntu 20.04 Found. Upgrading from GCC 9 to 10."
+        apt_install gcc-10 g++-10
+
+        echo_progress_start "Configuring GCC 10"
+        update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 10 >> $log 2>&1
+        update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-10 10 >> $log 2>&1
+        update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 9 >> $log 2>&1
+        update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 9 >> $log 2>&1
+        echo_progress_done "GCC 10 Configured"
+    else
+        echo_info "GCC version is 10. No upgrade required."
+    fi
+}
+
+Upgrade_GCC12_Jammy() {
+    local gcctwelve=$(gcc --version | grep -c "Ubuntu 12")
+    if [[ $gcctwelve -le 0 ]]; then
+        echo_info "Ubuntu 22.04 Found. Upgrading from GCC 11 to 12."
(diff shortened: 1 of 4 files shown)


sources/functions/gcc
```diff
@@ -0,0 +1,49 @@
...
+# GPLv3 Applies
+
+Upgrade_GCC_Ubuntu() {
+    echo_info "Checking OS for GCC upgrade"
+    if [[ "$(_os_codename)" = "focal" ]]; then
+        Upgrade_GCC10_Focal
+    elif [[ "$(_os_codename)" = "jammy" ]]; then
+        Upgrade_GCC12_Jammy
+    else
+        echo_info "Skipping GCC upgrade due to unsupported OS"
```
liaralabs, line 13: I don't see a reason to inform the user this was skipped. I would suggest an `echo_log_only` here

liaralabs:
I'm not fully against this idea, but overall, I think the quality of this code can be improved and made easier to maintain.

There's a lot of repeated code here and unnecessary function replication. The majority of this code is just duplicated and can be trimmed down to not have to copy paste entire blocks just to maintain.

Further, it would get major bonus points from me if the GCC version available can be automatically detected to avoid needing to maintain this script with every dist-upgrade.

liaralabs:
For example:

```
apt-cache search gcc | grep -oP '^gcc-\d?\d' | sort -Vr | uniq
```
Where the head is the most recent version

stickz:
We don't want to install the latest upstream version of `gcc` until it's stable enough. I will de-duplicate the entire upgrade routine on Friday. We can literally just take the new version and old version as parameters.

liaralabs:
The apt command is only pulling packages available to the OS. It won't determine GCC versions it doesn't have access to or aren't considered stable (otherwise why would a backport be available?)

If it's been released by upstream and it's stable and has made its way into Ubuntu packages on an LTS OS, I don't really see an issue with automating the version detection

stickz:
GCC is a special case. It's more complex than distributions. There are compiler regressions. We need to wait until at least one minor release, after every major release. We want to ensure stable gains in software performance are released to clients.

liaralabs:
I don't know if this is a concern worth worrying about though. GCC-13 is already being pushed in the short term releases and it's currently no where to be seen in the command output from above.

Further, GCC-12 was already on version 12.1 by the time Ubuntu Jammy even released in April 2022 (and it still is, they have not bumped to 12.2, despite it being released).

I don't think it's reasonable to expect that these packages will receive substantially different version upgrades than the rest of the "frozen" packages in the ubuntu stable repos. All fixes to this packages will be minor patch-type releases at the point of it being released

stickz:
The other concern is we need to test things. It's not a good idea to grab the latest version of GCC automatically. `rtorrent` could fail to compile for whatever reason. If we hardcore the new and previous version to a specific codename, it will reduce breakage. If the codename is not supported, it just won't upgrade. There is no mandatory maintenance involved.

liaralabs:
I've had some time to reflect on this as well. The proper place to push this GCC update would be the scripts/update/10-dependencies.sh file.

We either ship upgraded versions of GCC for all users on ubuntu or not at all. Since we install build-essential in this step, we should upgrade gcc here as well. This will automatically get pulled in for all ubuntu users the next time they run box update.

Having this just in the rtorrent script means it will only apply for certain users but not for others -- GCC is used elsewhere in the project, so we might run into edge cases where some folks on ubuntu have one version because they installed qbittorrent only and others installed rtorrent then qbittorrent which caused their qbit install to build with the differing version of GCC. I don't want to open us up to issues that might result from weird edge cases and tracking down niche/weird issues that result because users installed packages in different orders.

stickz:
I'm going to close this off and submit a new pull request. I don't like this way this one turned out.

PR #998, opened 2023-05-03
Nginx: Fix package errors during remove

stickz:
`php-mcrypt` and `php-geoip` are not supported on some operating systems. We need to check before attempting to remove them. Don't specify packages for removal more than once. If we miss anything, we can purge it. This is bad for maintenance.

### DIFF
--- a/scripts/remove/nginx.sh
+++ b/scripts/remove/nginx.sh
@@ -2,7 +2,8 @@
 
 systemctl stop -q nginx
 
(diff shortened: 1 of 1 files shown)


scripts/remove/nginx.sh
```diff
@@ -2,7 +2,24 @@
 
 systemctl stop -q nginx
 
-APT='nginx-extras nginx libnginx-mod-http-fancyindex ssl-cert php php-cli php-fpm php-dev php-xml php-curl php-xmlrpc php-json php-mcrypt php-opcache php-geoip php-xml php php-cli php-fpm php-dev php-xml php-curl php-xmlrpc php-json php-mcrypt php-opcache'
+codename=$(_os_codename)
+case $codename in
+    stretch)
```
liaralabs, line 7: Stretch and bionic are EOL which means they will never even see any updates beyond the last EOL commit designated in swizzin. These cases are basically cruft

flying-sausages:
@stickz did you test this on the systems it affects?

PR #1003, opened 2023-05-06
cURL: Optimized cmake building

stickz:
**Tested on Ubuntu 22.04 LTS ARM64**.
- Decreased build time from 3 minutes to 60 seconds on ARM64.
- Always enable Link Time Optimizations.
- Use Level 3 compiler optimizations with native march.


liaralabs:
Are you aware of any version requirements for cmake to build this? It might be a blocker on some of the older OSes but we do have a swizzin-cmake package available to help bypass this if it's a hurdle.

PR #1006, opened 2023-05-07
rTorrent: Consolidate package installs

stickz:
This commit consolidates package installs to be performed at the same time. It reduces install script overhead and makes it easier for the user to understand what is happening. Feature specific requirements are defined with a new variable, so developers know which ones to adjust if maintenance is required. Compatible with #1003.


liaralabs:
This doesn't really save any install time due to the way the apt_install function is programmed. This also further removes the dependencies from where they are actually used and outside of the functions which install them. Doing this is an anti-pattern when we rely self-contained functions to reduce unnecessary code replication.

PR #1007, opened 2023-05-07
Ubuntu: Upgrade to Python 3.11

stickz:
This pull request upgrades Ubuntu to Python 3.11, which is 60% faster than Python 3.10 for package installations. Changes were tested on Ubuntu 22.04 LTS ARM64. The installation speed of packages across the entire script was increased by 60%. This resolves a problem with slower deployment times on ARM, due to lengthy package installation times. x86 will benefit too.

### DIFF
--- a/scripts/update/10-dependencies.sh
+++ b/scripts/update/10-dependencies.sh
@@ -10,8 +10,11 @@ if [[ $(_os_distro) == "ubuntu" ]]; then
         echo_info "Upgrading to Python 3.11"
         add-apt-repository -y ppa:deadsnakes/ppa >> ${log} 2>&1
         apt_install python3.11-full
-        echo "alias python=/usr/bin/python3.11" >> ~/.bashrc
-        echo "alias python3=/usr/bin/python3.11" >> ~/.bashrc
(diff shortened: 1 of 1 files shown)


liaralabs:
As a general rule, it is to be avoided changing the distribution version of python away from defaults. We use virtualenvs to ship versions of python which aren't the distro default.

stickz:
This is worth changing because it takes minutes off the installation process when lots of packages are installed. As stated, you can just use a virtualenv in the unlikely event it breaks anything. This version bump is 60% faster than the predecessor. It makes the installation script rapid and viable on ARM platforms; which especially suffer some terrible package installation times. I don't think we should hold back a package upgrade like this, when we can make everything we want compatible using virtualenvs.

brettpetch:
We don’t change system Python.

brettpetch:
Use pyenv. It’s in sources/functions/pyenv

stickz:
I'll just close it off if you don't want it. It's not possible to use a python environment for the `apt package manager`. Better off to **double** the time it takes to install packages for the installation script.

PR #1009, opened 2023-05-10
Jammy: GCC 12 system default

stickz:
This commit sets GCC 12 as the Ubuntu 22.04 LTS system default. It allows users to unlock the latest improvements for compiled software solutions. It is capable of repairing linkage from previous distribution releases or manual user adjustments.

Implemented as suggested in #992. The plan is to only support Jammy for simplicity. It is pointless upgrading Focal because it's already two versions behind. The user should upgrade their distribution release, if they want the latest software packages.

Tested on Ubuntu 22.04 LTS ARM64. The script was successful.

### DIFF
--- a/scripts/update/10-dependencies.sh
+++ b/scripts/update/10-dependencies.sh
@@ -7,7 +7,7 @@ fi
 
 if [[ $(_os_distro) == "ubuntu" ]]; then
     if [[ $(_os_codename) == "jammy" ]]; then
-        if ! grep 'ubuntu-toolchain-r' /etc/apt/sources.list | grep -q -v '^#'; then
+        if ! grep 'ubuntu-toolchain-r' /etc/apt/sources.list.d/ubuntu-toolchain-r-ubuntu-ppa-jammy.list | grep -q -v '^#'; then
             echo_info "Adding toolchain repo"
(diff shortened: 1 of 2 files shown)


stickz:
Pull request ready for review again. GCC `12.3` successfully deployed on Ubuntu 22.04 when testing on ARM64.
```
:~# gcc --version
gcc (Ubuntu 12.3.0-1ubuntu1~22.04) 12.3.0
Copyright (C) 2022 Free Software Foundation, Inc.
```

scripts/update/10-dependencies.sh
```diff
@@ -6,6 +6,13 @@ if ! which add-apt-repository > /dev/null; then
 fi
 
 if [[ $(_os_distro) == "ubuntu" ]]; then
+    if [[ $(_os_codename) == "jammy" ]]; then
+        if ! grep 'ubuntu-toolchain-r' /etc/apt/sources.list | grep -q -v '^#'; then
```
liaralabs, line 10: This is an invalid check. On jammy, the repo is configured to 

`/etc/apt/sources.list.d/ubuntu-toolchain-r-ubuntu-ppa-jammy.list`

PR #1011, opened 2023-05-10
Nginx: dhparm background task

stickz:
This commit starts a background task for `openssl dhparam -out dhparam.pem` to allow package installation at the same time. This reduces installation time of `nginx` by 30s on x86 and 60s on ARM64 without any real disadvantages. `temp.log` is created to prevent a race condition with logging. The logging output is appended to `swizzin.log` afterwards and `temp.log` is removed.

Tested on Ubuntu 22.04 LTS ARM64. Works as expected. Package installation finished before `openssl dhparam -out dhparam.pem`. The bash script waited for the background task to finish before proceeding. 60s of installation time was saved.

### DIFF
--- a/scripts/install/nginx.sh
+++ b/scripts/install/nginx.sh
@@ -51,11 +51,35 @@ case $codename in
         ;;
 esac
 
+# Prepare the /etc/nginx/ssl/ for openssl dhparm generation
+mkdir -p /etc/nginx/ssl/
+chmod 700 /etc/nginx/ssl
+cd /etc/nginx/ssl
+
+# Create temp.log for openssl dhparm generation to prevent a race condition with logging
+. /etc/swizzin/sources/functions/utils
+templog="/root/logs/temp.log"
+rm_if_exists $templog
+touch $templog
+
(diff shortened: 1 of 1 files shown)


flying-sausages:
I'm actually ok with this one in principle, @liaralabs thoughts?

scripts/install/nginx.sh
```diff
@@ -51,11 +51,35 @@ case $codename in
...
+# Install packages for nginx in the foreground
+APT="nginx libnginx-mod-http-fancyindex subversion ssl-cert php-fpm libfcgi0ldbl php-cli php-dev php-xml php-curl php-xmlrpc php-json php-mbstring php-opcache php-zip ${geoip} ${mcrypt}"
 apt_install $APT
-mkdir -p /srv
 
+# Wait for the background task of openssl dhparm generation to finish
+wait
+
+# Append the results of temp.log to swizzin.log and remove temp.log
+echo -e "\nBegin of OpenSSL dhparm results" >> $log 2>&1
```
liaralabs, line 76: We have an echo_log_only function that would be a better fit here

PR #1015, opened 2023-05-21
rTorrent: Network configuration

stickz:
This commit implements a 4MB send and receive buffer for TCP sockets. It improves throughput on rTorrent. Debian/Ubuntu will automatically downscale this value if the user doesn't have enough memory. The maximum default without changes is 4MB.

It also adds a missing configuration variable for per torrent upload slots. The default value of 50 is applied.

### DIFF
--- a/scripts/install/rtorrent.sh
+++ b/scripts/install/rtorrent.sh
@@ -24,6 +24,8 @@ network.port_range.set = $port-$portend
 network.scgi.open_local = /var/run/${user}/.rtorrent.sock
 schedule2 = chmod_scgi_socket, 0, 0, "execute2=chmod,\"g+w,o=\",/var/run/${user}/.rtorrent.sock"
 network.tos.set = throughput
+network.send_buffer.size.set = 4M
+network.receive_buffer.size.set = 4M
 pieces.hash.on_completion.set = no
 pieces.preload.min_rate.set = 50000
 protocol.pex.set = no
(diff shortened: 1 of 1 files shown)


liaralabs:
This is left unconfigured on purpose so that rtorrent falls back to the linux kernel settings instead

stickz:
> This is left unconfigured on purpose so that rtorrent falls back to the linux kernel settings instead

Without this setting, it it will use something silly like 4KB. I tested this at 2gbps. There was a lot of upstream instability. When I increased it to 4MB, I was able to sustain my upload speed properly. It's automatically scaled back without sufficient memory.

liaralabs:
It will fall back to the values of rmem and wmem as configured in the kernel, ie:

```
net.ipv4.tcp_rmem
net.ipv4.tcp_wmem
```

If people set these during tuning (very common) I don't want the defaults of rtorrent to override the kernel level settings. Relying on the linux kernel over specific application settings is preferable.

stickz:
> If people set these during tuning (very common) I don't want the defaults of rtorrent to override the kernel level settings. Relying on the linux kernel over specific application settings is preferable.

I disagree that we should prefer tweaking the Linux Kernel settings to make a software application work properly. The vast majority of people expect it to work after installation. We could allow up to 3gbps on a single socket with no tweaking required.

This setting won't go over the specified maximum values as configured in the kernel. It's perfectly safe to perform this operation.
```
net.core.rmem_max
net.core.wmem_max
```

brettpetch:
We didn’t imply that we should be tweaking them at all.

stickz:
> We didn’t imply that we should be tweaking them at all. That’s more up to the user than anyone.

We should at least provide them in the configuration file, explain what they do and comment them off so they don't apply by default. We shouldn't expect users to research and diagnose why their upload speed is very unstable on `rTorrent`. This is like saying you're not going to include the upload slot count in the configuration file because you "shouldn't be tweaking it".

flying-sausages:
I wouldn't be too opposed to more comments in that config file, and some additional commented suggestions to be honest. If you feel like doing that, feel free to to push that and I will re-open this for discussion

PR #1016, opened 2023-05-22
rTorrent: Network configuration suggestion

stickz:
This commit suggests implementation of a 4MB send and receive buffer for TCP sockets to improve throughput on gigabit connections. It explains why they are disabled by default on rTorrent and how much memory is consumed.

It also adds a missing configuration variable for per torrent upload slots. The default value of 50 is applied.

### DIFF
--- a/scripts/install/rtorrent.sh
+++ b/scripts/install/rtorrent.sh
@@ -33,10 +33,20 @@ throttle.global_down.max_rate.set = 0
 throttle.global_up.max_rate.set = 0
 throttle.max_peers.normal.set = 100
 throttle.max_peers.seed.set = -1
+throttle.max_uploads.set = 50
 throttle.max_uploads.global.set = 100
 throttle.min_peers.normal.set = 1
 throttle.min_peers.seed.set = -1
 trackers.use_udp.set = yes
+
+# Disabled by default becuase these will override Linux Kernel settings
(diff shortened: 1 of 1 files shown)


scripts/install/rtorrent.sh
```diff
@@ -33,11 +33,21 @@ throttle.global_down.max_rate.set = 0
...
 throttle.max_peers.normal.set = 100
 throttle.max_peers.seed.set = -1
+throttle.max_uploads.set = 50
 throttle.max_uploads.global.set = 100
 throttle.min_peers.normal.set = 1
 throttle.min_peers.seed.set = -1
 trackers.use_udp.set = yes
 
+# Disabled by default becuase these will override Linux Kernel settings
+# Suggested to configure on a per application basis to improve throughput
```
liaralabs, line 43: Remove 43-46 and link to documentation instead.

Also this is not my suggestion nor the opinion of other maintainers as many users have server oriented kernel configs installed during their setup.

I would allow wording closer to "you may wish to configure..." In the docs article that will be linked to, so as to not potentially confuse users into thinking they blindingly need to apply configs they may not understand but could potentially override other pre-configured defaults as shipped by their provider.

liaralabs (changes requested):

stickz:
This pull request is marked a draft for now. I will return it when time permits.

PR #1032, opened 2023-07-04
Fix Jellyfin not installing correctly

KatieTheDev:
## Description

Jellyfin split their repositories into two and have explained why [here](https://jellyfin.org/docs/general/installation/linux/#ubuntu). To accommodate this change, I have updated the installation script and update script to accommodate their new repository changes.

## Fixes issues: 
- Un-tracked issue: Jellyfin stopped installing, giving an error that either a repository was not signed or completely refusing to connect.

## Proposed Changes:
- Changed the installation script to install the latest signing key.
- Changed the installation script to add the new repository depending on the OS and version the user is running.
- Changed the update script to remove old signing keys because a previous signing key has expired.
- Changed the update script to remove old repositories because the repository has changed.

## Change Categories

- Bug fix

## Checklist

    - PR link: Doc changes unnecessary
    - PR link: Panel changes unnecessary
- [x] Testing was done
   - [x] Tests executed

## Test scenarios

I tested by running box install and box update using these new scripts. The embedded video was created before I finished work on the update script, but demonstrates how both scripts function. I also did more commits to the install script after making the video, but everything worked anyway as these were mainly stylistic changes or changes to echoes. Test performed on Ubuntu Server 22.04.2 VM in Proxmox VE on a Dell Poweredge R720xd running a fresh install of Swizzin. Only other packages installed on Swizzin were panel and nginx.

If you want to watch the test video in 1080p:
https://youtu.be/Swa9ESljrVU

Uploaded in 480p to meet GitHub size limits:
https://github.com/swizzin/swizzin/assets/51092829/d0f279c7-0c6d-41a6-8207-c5c5addf1726

### Architectures

| Jammy 		|	✅		|			|			|				|
| Focal 		|			|			|			|				|
| Buster		|			|			|			|				|

### DIFF
--- a/scripts/install/jellyfin.sh
+++ b/scripts/install/jellyfin.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# authors: liara userdocs flying-sausages
+# authors: liara userdocs flying-sausages katiethedev
 #
 # Licensed under GNU General Public License v3.0 GPL-3 (in short)
 #
@@ -39,6 +39,83 @@ if [[ -n $active ]]; then
         exit 1
     fi
 fi
+
+ARCHITECTURE="$(dpkg --print-architecture)"
+BASE_OS="$(awk -F'=' '/^ID=/{ print $NF }' /etc/os-release)"
+
+# Validate that we're running on a supported (dpkg) architecture
+# shellcheck disable=SC2254
+# We cannot quote this extglob expansion or it doesn't work
+case "${ARCHITECTURE}" in
+    ${SUPPORTED_ARCHITECTURES})
+        true
+        ;;
+    *)
+        echo "Sorry, Jellyfin doesn't support the CPU architecture '${ARCHITECTURE}'."
+        exit 1
+        ;;
+esac
+
+# Handle some known alternative base OS values with 1-to-1 mappings
+# Use the result as the repository base OS
+case "${BASE_OS}" in
+    raspbian)
+        # Raspbian uses the Debian repository
+        REPO_OS="debian"
+        ;;
+    linuxmint)
+        # Linux Mint can either be Debian- or Ubuntu-based, so pick the right one
+        if grep -q "DEBIAN_CODENAME=" /etc/os-release &> /dev/null; then
+            VERSION="$(awk -F'=' '/^DEBIAN_CODENAME=/{ print $NF }' /etc/os-release)"
+            REPO_OS="debian"
+        else
+            VERSION="$(awk -F'=' '/^UBUNTU_CODENAME=/{ print $NF }' /etc/os-release)"
+            REPO_OS="ubuntu"
+        fi
+        ;;
+    neon)
+        # Neon uses the Ubuntu repository
+        REPO_OS="ubuntu"
+        ;;
+    *)
+        REPO_OS="${BASE_OS}"
+        VERSION="$(awk -F'=' '/^VERSION_CODENAME=/{ print $NF }' /etc/os-release)"
+        ;;
+esac
+
+# Validate that we're running a supported release (variables at top of file)
+case "${REPO_OS}" in
+    debian)
+        # shellcheck disable=SC2254
+        # We cannot quote this extglob expansion or it doesn't work
+        case "${VERSION}" in
+            ${SUPPORTED_DEBIAN_RELEASES})
+                true
+                ;;
+            *)
+                echo "Sorry, we don't support the Debian codename '${VERSION}'."
+                exit 1
+                ;;
+        esac
+        ;;
+    ubuntu)
+        # shellcheck disable=SC2254
+        # We cannot quote this extglob expansion or it doesn't work
+        case "${VERSION}" in
+            ${SUPPORTED_UBUNTU_RELEASES})
+                true
+                ;;
+            *)
+                echo "Sorry, we don't support the Ubuntu codename '${VERSION}'."
+                exit 1
+                ;;
+        esac
+        ;;
+    *)
+        echo "Sorry, we don't support the base OS '${REPO_OS}'."
+        exit 1
+        ;;
+esac
 
 #
 ########
@@ -88,7 +165,7 @@ CONFIG
 cat > /etc/jellyfin/network.xml <<- CONFIG
 <?xml version="1.0" encoding="utf-8"?>
 <NetworkConfiguration xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
-  <RequireHttps>true</RequireHttps>
+  <RequireHttps>false</RequireHttps>
   <CertificatePath>/home/${username}/.ssl/${username}-self-signed.pfx</CertificatePath>
   <CertificatePassword />
   <BaseUrl />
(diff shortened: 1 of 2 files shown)


scripts/install/jellyfin.sh
```diff
@@ -40,6 +34,47 @@ if [[ -n $active ]]; then
     fi
 fi
 
+ARCHITECTURE="$(dpkg --print-architecture)"
```
liaralabs, line 37: We have a builtin for this: `_os_arch`

scripts/install/jellyfin.sh
```diff
@@ -40,6 +34,47 @@ if [[ -n $active ]]; then
...
+BASE_OS="$(awk -F'=' '/^ID=/{ print $NF }' /etc/os-release)"
```
liaralabs, line 38: `_os_distro`

scripts/install/jellyfin.sh
```diff
@@ -40,6 +34,47 @@ if [[ -n $active ]]; then
...
+            REPO_OS="ubuntu"
+        fi
+        ;;
+    neon)
+        # Neon uses the Ubuntu repository
+        REPO_OS="ubuntu"
+        ;;
+    *)
+        REPO_OS="${BASE_OS}"
+        VERSION="$(awk -F'=' '/^VERSION_CODENAME=/{ print $NF }' /etc/os-release)"
```
liaralabs, line 63: `_os_codename`

scripts/install/jellyfin.sh
```diff
@@ -40,6 +34,47 @@ if [[ -n $active ]]; then
     fi
 fi
 
+ARCHITECTURE="$(dpkg --print-architecture)"
+BASE_OS="$(awk -F'=' '/^ID=/{ print $NF }' /etc/os-release)"
+
+# Handle some known alternative base OS values with 1-to-1 mappings
+# Use the result as the repository base OS
+case "${BASE_OS}" in
```
liaralabs, line 42: This case is basically useless aside from setting the _os_codename. If an OS doesn't directly identify itself as Debian or Ubuntu (i.e. during install) then we don't support it.

We certainly don't support linux mint or kde neon

scripts/install/jellyfin.sh
```diff
@@ -88,7 +123,7 @@ CONFIG
 cat > /etc/jellyfin/network.xml <<- CONFIG
 <?xml version="1.0" encoding="utf-8"?>
 <NetworkConfiguration xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
-  <RequireHttps>true</RequireHttps>
+  <RequireHttps>false</RequireHttps>
```
liaralabs, line 126: The original spirit of this script created self-signed certs for use with Jellyfin, which is why this was left enabled.

scripts/update/jellyfin.sh
```diff
@@ -68,15 +62,79 @@ if [[ -f /install/.jellyfin.lock ]]; then
...
     #
     if ! check_installed jellyfin; then
-        echo_info "Moving Jellyfin to apt-managed installation"
+        echo_info "Updating Jellyfin installation using apt."
         #
-        # Add the jellyfin official repository and key to our installation so we can use apt-get to install it jellyfin and jellyfin-ffmepg.
-        curl -s https://repo.jellyfin.org/$DIST_ID/jellyfin_team.gpg.key | gpg --dearmor > /usr/share/keyrings/jellyfin-archive-keyring.gpg 2>> "${log}"
-        echo "deb [signed-by=/usr/share/keyrings/jellyfin-archive-keyring.gpg arch=$(dpkg --print-architecture)] https://repo.jellyfin.org/$DIST_ID $DIST_CODENAME main" > /etc/apt/sources.list.d/jellyfin.list
+        # Make sure universe is enabled so that ffmpeg can be satisfied.
+        sudo add-apt-repository universe
```
liaralabs, line 68: This is ubuntu-centric code that doesn't work on Debian.

Also, all Ubuntu repos are guaranteed to be installed:

https://github.com/swizzin/swizzin/blob/master/scripts/update/10-dependencies.sh#L8-L46

scripts/update/jellyfin.sh
```diff
@@ -68,15 +62,79 @@ if [[ -f /install/.jellyfin.lock ]]; then
...
+
+        # Handle some known alternative base OS values with 1-to-1 mappings
+        # Use the result as the repository base OS
+        ARCHITECTURE="$(dpkg --print-architecture)"
```
liaralabs, line 72: Use built-ins

scripts/update/jellyfin.sh
```diff
@@ -68,15 +62,79 @@ if [[ -f /install/.jellyfin.lock ]]; then
...
+        BASE_OS="$(awk -F'=' '/^ID=/{ print $NF }' /etc/os-release)"
```
liaralabs, line 73: Use built-ins

scripts/update/jellyfin.sh
```diff
@@ -68,15 +62,79 @@ if [[ -f /install/.jellyfin.lock ]]; then
...
+        case "${BASE_OS}" in
```
liaralabs, line 74: replicated code should be a function; however, this is useless code beyond the setting of the distro codename (as above)

scripts/update/jellyfin.sh
```diff
@@ -68,15 +62,79 @@ if [[ -f /install/.jellyfin.lock ]]; then
...
+        # Add Jellyfin signing key if not already present
+        if [[ ! -f /etc/apt/keyrings/jellyfin.gpg ]]; then
+            echo_progress_start "> Did not find signing key. Adding it."
+            curl -fsSL https://repo.jellyfin.org/jellyfin_team.gpg.key | gpg --dearmor --yes --output /etc/apt/keyrings/jellyfin.gpg
+            echo_progress_done "Jellyfin Signing Key Added"
+        fi
+
+        #
+        # Install the Deb822 format jellyfin.sources entry
+        echo_progress_start "Adding Jellyfin repository to apt"
```
liaralabs, line 119: Replicated code should be functionized (sources/functions/jellyfin), however I wouldn't block merging based on this, but it would be best-practices for long-term ease of maintenace.

liaralabs (changes requested):
Thanks for the PR KatieTheDev. There are a few issues that cropped up during the review that should be resolved, but this otherwise looks good.

Thanks for the contribution!

KatieTheDev:
I will make the desired changes. Also, I realized that `universe` was redundant at some point in revising before I submitted the PR, but I must’ve forgotten to either fix that or commit the fix. I also didn’t know about the builtins you mentioned. Also, the reason I disabled force https is because I believe it breaks app support. I will try again with it enabled and see if there’s another workaround for that bug. Something is causing the app not to see the server in the default configuration. I think it may be binding to 127.0.0.1 when nginx is enabled which is blocking outside connections and the apps don’t really like going through nginx. I’ll spend some time on this when I can.

scripts/install/jellyfin.sh
KatieTheDev: I was unaware we don’t support those distros at all. I will be removing them from the scripts.

scripts/install/jellyfin.sh
flying-sausages:
```suggestion
ARCHITECTURE="$(_os_arch)"
```

scripts/install/jellyfin.sh
flying-sausages:
```suggestion
BASE_OS="$(_os_distro)"
```

scripts/install/jellyfin.sh
flying-sausages:
```suggestion
        VERSION="$(_os_codename)"
```

scripts/update/jellyfin.sh
flying-sausages:
```suggestion
        ARCHITECTURE="$(_os_arch)"
```

scripts/update/jellyfin.sh
flying-sausages:
```suggestion
        BASE_OS="$(_os_distro)"
```

flying-sausages:
> Also, the reason I disabled force https is because I believe it breaks app support. I will try again with it enabled and see if there’s another workaround for that bug. Something is causing the app not to see the server in the default configuration. I think it may be binding to 127.0.0.1 when nginx is enabled which is blocking outside connections and the apps don’t really like going through nginx. I’ll spend some time on this when I can.

like mobile apps? If that were the case we would have had a lot more people complain about that by now I think

liaralabs:
I do think some of the mobile apps won't work with self-signed certificates. The use-case of this self-signed cert is limited only to installs without the nginx proxy though, so the amount of people using self-signed certs on installs without nginx is probably limited. That said, it would also affect folks without a domain/let's encrypt

KatieTheDev:
Would it be acceptable for me to add instructions for turning it off in the wiki as a potential fix for app issues?

flying-sausages:
> Would it be acceptable for me to add instructions for turning it off in the wiki as a potential fix for app issues?

That, plus maybe a `echo_docs` in the code for when you can see it might be relevant to the users maybe, but not a required addition from my end

KatieTheDev:
I don't really know how to fix the conflicts that need to be resolved.

KatieTheDev:
I believe that I have resolved merge conflicts and I have merged the latest updates into my branch. I believe I have also resolved the blocking changes, as I have switched everything to builtins that I noticed in your suggestions. I will now work on functionizing things.

scripts/install/jellyfin.sh
KatieTheDev: I removed unsupported distros, but I am not aware if _os_codename is set as debian on raspbian. It is necessary to make sure raspbian is set to the same repos as debian, as that is what Jellyfin requires.

KatieTheDev:
I am unsure why pre-commit is showing a failure. I have no errors in shellcheck.

scripts/update/jellyfin.sh
KatieTheDev: I will use _os_codename in both cases if you can confirm that raspbian is set to debian. If not, I will create a function.

KatieTheDev:
I added a new commit splitting the repository into a separate function. I was also able to include those new variables into it, as the only place those were used was in that function.

scripts/install/jellyfin.sh
```diff
@@ -40,6 +34,17 @@ if [[ -n $active ]]; then
     fi
 fi
 
+#
+## Get the path to gpg or install it
+GNUPG=$(which gpg)
```
liaralabs, line 39: gnupg2 is a global dependency of swizzin. You can safely assume every user has this installed already.

sources/functions/jellyfin
```diff
@@ -0,0 +1,32 @@
...
+# Contributers: katiethedev
+# This functions file is created to resolve some replicated code between the install and update functions.
+
+# This function adds the Jellyfin repository.
+function add_jellyfin_repo() {
+    # Set required variables
+    ARCHITECTURE="$(_os_arch)"
+    BASE_OS="$(_os_distro)"
+    case "${BASE_OS}" in
+        raspbian)
```
liaralabs, line 11: I don't think raspbian will pass [this](https://github.com/swizzin/swizzin/blob/develop/setup.sh#L197) check so we don't need to write a case for it

liaralabs:
LGTM other than the two comments

liaralabs:
> I am unsure why pre-commit is showing a failure. I have no errors in shellcheck.

I believe the errors were a result of sausage doing the dev work on pre-commit checks at the same time and the pipeline issues have been resolved. Next time you commit it should re-check :crossed_fingers:

PR #1039, opened 2023-07-14
GitHub tools

flying-sausages:
## Description
Standardise use of the github api for querying for download URLs and versions

## Proposed Changes:
- New function for getting download URLs which also exports the retrieved tag as a variable

## Change Categories

- Change in `functions` 

## Checklist

    - PR link: 
    - PR link: 
- [ ] Testing was done
   - [ ] Tests executed

## Test scenarios

Should test that all affected installers work as expected

### Architectures

| Jammy 		|			|			|			|				|
| Focal 		|			|			|			|				|
| Buster		|			|			|			|				|

### DIFF
--- a/scripts/upgrade/sabnzbd.sh
+++ b/scripts/upgrade/sabnzbd.sh
@@ -10,15 +10,11 @@ fi
 . /etc/swizzin/sources/functions/pyenv
 localversion=$(/opt/.venv/sabnzbd/bin/python /opt/sabnzbd/SABnzbd.py --version | grep -m1 SABnzbd | cut -d- -f2)
 #latest=$(curl -s https://sabnzbd.org/downloads | grep -m1 Linux | grep download-link-src | grep -oP "href=\"\K[^\"]+")
-latestversion=$(github_latest_version sabnzbd/sabnzbd) || {
-    echo_error "Failed to query GitHub for latest sabnzbd version"
-    exit 1
-}
-latest="https://github.com/sabnzbd/sabnzbd/archive/refs/tags/${latestversion}.tar.gz"
+latest=$(github_release_url sabnzbd/sabnzbd "src.tar.gz")
 
 pyvenv_version=$(/opt/.venv/sabnzbd/bin/python --version | awk '{print $2}')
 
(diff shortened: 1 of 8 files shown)


liaralabs:
Before you invest a lot of time in this, the whole point of the github latest version function currently in use is precisely that it _doesn't_ use the github api because the rate limits can be avoided entirely this way

flying-sausages:
ok 60/hour is probably something that is indeed going to bite us if we use the API.

Yeah changing the guts of the function to one that uses the "scraper" could be beneficial so we get rid of the API use in the rest of the codebase easier as well, and can then just adapt our approach in one swoop in case there's a change in how the release page looks.

Could just hop onto your latest tag function and then return a formatted string for the download URL

How do you feel about exporting the tag version as a variable as a byproduct of this function? It smells like an anti-pattern to me

flying-sausages:
Re-wrote this without using the API, and found the endpoint Github uses to generate the list of releases that's used by their frontend so we can parse it a lot easier as well.

The code is a lot cleaner now too, but I'm still unsure about the version variable. I'm thinking we can just expect that you get the version first if you need it as a variable, and you can pass it into this function so you can skip the call as well?

scripts/upgrade/sabnzbd.sh
```diff
@@ -10,15 +10,11 @@ fi
...
-    echo_error "Failed to query GitHub for latest sabnzbd version"
-    exit 1
-}
-latest="https://github.com/sabnzbd/sabnzbd/archive/refs/tags/${latestversion}.tar.gz"
+latest=$(github_release_url sabnzbd/sabnzbd "src.tar.gz")
 
 pyvenv_version=$(/opt/.venv/sabnzbd/bin/python --version | awk '{print $2}')
 
-if dpkg --compare-versions ${pyvenv_version} lt 3.8.0 && dpkg --compare-versions ${latestversion} ge 3.5.0; then
+if dpkg --compare-versions ${pyvenv_version} lt 3.8.0 && dpkg --compare-versions ${github_sabnzbd_version} ge 3.5.0; then
```
liaralabs, line 17: ${github_sabnzbd_version} is undefined

PR #1066, opened 2023-09-03
rTorrent: Bump xmlrpc-c to super stable

stickz:
Resolves: `svn: E155000: '/tmp/xmlrpc-c' is already a working copy for a different URL`

I came across this error when testing the develop branch. The first instance of `svn co` failed, which caused the second one to fail. The simplest solution is bump xmlrpc-c to super-stable and update the link to the svn handle.


liaralabs:
Github is sunsetting svn support in the coming months, so we need to decrease reliance on git svn hacks instead of increase it:

https://github.blog/2023-01-20-sunsetting-subversion-support/

Issue #1071, opened 2023-09-08
[BUG] Calibreweb Nginx configuration doesn't work for IPv6


Irishsmurf:
#### Steps to reproduce:

1. Install nginx.
   * `box install nginx`
2. Install CalibreWeb
    * `box install calibreweb`  
3. Attempt to load `https://<website>/calibreweb` from an IPv6 Address
5. `500 Internal Error`
    * From Nginx Error logs: 
    * ```
       2023/09/08 16:45:16 [crit] 2382422#2382422: *490 bind(<IPv6 IP ADDRESS>) failed (97: Address family not supported by 
       protocol) while connecting to upstream, client: <IPv6 IP ADDRESS>, server: <SERVER HOSTNAME>, 
       request: "GET /calibreweb HTTP/2.0", upstream: "http://127.0.0.1:8083/calibreweb", host: "<SERVER HOSTNAME>", referrer: 
       "https://<SERVER HOSTNAME>/"
       ```

#### Expected results:

Page loads.

#### Results occur:

CalibreWeb is inaccessible if the client is using IPv6.

---

I fixed this by modifying the nginx configuration in `calibreweb.conf` to:

```conf
location /calibreweb {
        proxy_pass              http://127.0.0.1:8083;
        proxy_set_header        Host $proxy_host;
        proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header        X-Forwarded-Host $host;
        proxy_set_header        X-Forwarded-Proto $scheme;
        proxy_http_version      1.1;
        proxy_set_header        Upgrade $http_upgrade;
        proxy_set_header        Connection $http_connection;

        proxy_set_header        X-Scheme        $scheme;
        proxy_set_header        X-Script-Name   /calibreweb;  # IMPORTANT: path has NO trailing slash
}
```

### Swizzin commit

9dc0c50

### What OS are you using?

Debian 10 (Buster)

### What architecture is your OS?

amd64

### Relevant logs and output

```plaintext
2023/09/08 16:45:16 [crit] 2382422#2382422: *490 bind(<IPv6 IP ADDRESS>) failed (97: Address family not supported by 
       protocol) while connecting to upstream, client: <IPv6 IP ADDRESS>, server: <SERVER HOSTNAME>, 
       request: "GET /calibreweb HTTP/2.0", upstream: "http://127.0.0.1:8083/calibreweb", host: "<SERVER HOSTNAME>", referrer: 
[...]

PR #1072, opened 2023-09-10
Don't set `proxy_bind` and instead rely on default behaviour.

Irishsmurf:
Setting `proxy_bind` to `$server_addr` can cause issues if the connecting client is attempting to connect to a different IP protocol address than the backend supports (IPv6 vs IPv4)

This removes explicitly setting the `proxy_bind` setting, and instead relies on the default nginx behaviour (of binding to the same protocol as the upstream service)

## Description

`calibreweb` currently doesn't work with IPv6 as internally, CalibreWeb binds to an IPv4 address `127.0.0.1` - and if you attempt to hit Nginx with a client using IPv6 - `$server_addr` will be an IPv6 address - preventing it to talk to the upstream IPv4 address

## Fixes issues: 
- Issue #1071 1071

## Proposed Changes:
- Changing `/scripts/nginx/calibreweb.sh` to remove setting `proxy_bind`, and instead rely on default behaviour.

## Change Categories

- Bug fix               

## Checklist

- [✅] Branch was made off the `develop` branch and the PR is targetting the `develop` branch
- [✅] Docs have been made OR are not necessary
- [✅] Changes to panel have been made OR are not necessary
- [✅] Code conforms to project structure [(See more)](https://swizzin.ltd/dev/structure)
- [✅] Prints to terminal are handled [(See more)](https://github.com/swizzin/swizzin/blob/master/CONTRIBUTING.md#printing-into-the-terminal)
- [✅] I have commented my code, particularly in hard-to-understand areas
- [✅] Testing was done
   - [✅] Tests created or no new tests necessary
   - [✅] Tests executed

## Test scenarios

I manually modified the Nginx Configuration.

1. Without the change, it errors with `500 Internal Error`
2. With the change, the page loads over both IPv4 and IPv6

### Architectures

| Jammy 		|			|			|			|				|
| Focal 		|			|			|			|				|
| Buster		|			|			|			|				|

### DIFF
--- a/scripts/nginx/calibreweb.sh
+++ b/scripts/nginx/calibreweb.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 cat > /etc/nginx/apps/calibreweb.conf << EOF
 location /calibreweb {
-        proxy_bind              \$server_addr;
         proxy_pass              http://127.0.0.1:8083;
         proxy_set_header        Host            \$http_host;
         proxy_set_header        X-Forwarded-For \$proxy_add_x_forwarded_for;
--- a/scripts/update/calibreweb.sh
+++ b/scripts/update/calibreweb.sh
@@ -16,3 +16,13 @@ if [[ -f /install/.calibreweb.lock ]]; then
         echo_progress_done
     fi
 fi
+
+# Remove the proxy_bind setting from Nginx Conf.
+if [[ -f /etc/nginx/apps/calibreweb.conf ]]; then
+    if grep -q "proxy_bind[ \t]\+\$server_addr;" /etc/nginx/apps/calibreweb.conf ]]; then
+        echo_log_only "Removing proxy_bind from CalibreWeb nginx conf"
+        # Find the proxy_bind line, and remove it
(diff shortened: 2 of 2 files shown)


liaralabs:
Thanks for doing all the legwork on this.

Can you please supply a script for the update directory and add a case to the existing calibreweb.sh?

I'll leave the exact mechanism up to you, but the gist is that we should check wither proxy_bind is in the nginx conf file and set the config correctly.

A simple mechanism would be to simply trigger the nginx/calibreweb.sh script as it should do what we need to already

Irishsmurf:
Hey,

I've updated `calibreweb.sh` which adds a case which will overwrite an existing nginx config for CalibreWeb with one which removes the `proxy_bind` setting.

I took inspiration from `mango.sh` - which also overwrites an existing nginx config.

Is there something specific that's needed to be done to trigger this, or will this just run when calling `box update`? 
I took a look at the update directory and it seems everything is merely tied to an app-specific shell script - and there's no triggers that I can see outside of that.

liaralabs:
Yeah, you don't need to worry about that -- the update/upgrade hooks run all scripts in the update directory so nothing further than writing the hook is involved.

PR #1113, opened 2024-01-01
Changed boost download

siedenburg2:
Swapped boost download url because the account on the main site is deactivated, also changed wget to curl to download more than an html file from sourceforge

## Description

The boost download redirects to landing.jfrog.com/reactivate-server/boostorg, wget loads a html page and because it downloads something the fallback isn't working, could be fixed, but is a bit more work, so I canged the dl sites for now as a quick fix. Also changed wget to curl because I got problems with wget and loading from sourceforge.

Also: Is there a reason why boost 1.75 is still in use while there was a boost 1.84 release this december? 1.75 was released in december 2020. Didn't change version in case of breaking stuff.

## Fixes issues: 
- Un-tracked issue...

## Proposed Changes:
- Changed boost download method in libtorrent function

## Change Categories

- Bug fix               
- Change in `functions` 

## Checklist

    - PR link: 
    - PR link: 
- [x] Testing was done
   - [x] Tests executed

### Architectures

| Jammy 		|	✅		|			|			|				|
| Focal 		|			|			|			|				|
| Bookworm      |    ✅       |           |           |               |
| Buster		|			|			|			|				|

### DIFF
--- a/sources/functions/libtorrent
+++ b/sources/functions/libtorrent
@@ -148,13 +148,41 @@ function booststrap() {
     if [[ ! -d ${BOOST_ROOT} ]]; then
         echo_progress_start "Installing boost"
         cd /opt
-        curl -O https://deac-ams.dl.sourceforge.net/project/boost/boost/${BOOST_VERSION//_/.}/boost_${BOOST_VERSION}.tar.gz >> ${log} 2>&1 || {
-            echo_warn "Could not download boost from sourceforge mirror, attempting fallback."
-            wget https://boostorg.jfrog.io/artifactory/main/release/${BOOST_VERSION//_/.}/source/boost_${BOOST_VERSION}.tar.gz >> ${log} 2>&1 || {
-                echo_error "Boost could not be downloaded. Setup will exit."
-                exit 1
-            }
-        }
+        declare -A boost_mirror_urls=(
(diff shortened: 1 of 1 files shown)


liaralabs:
I think attempting to make the fallback mechanism robust would be the preferred play here. Sourceforge isn't the primary mirror and I'm not interested in swapping the mirrors to reflect that.

Reason being, the upstream issues have already been resolved without intervention from us. Making snap decisions such as this to swap the mirrors does fix the issue in the interim but doesn't actually address the real issue here: fallback was temporarily needed and didn't properly trigger due to the command not actually failing (receiving a redirect instead of a download still give us an exit code of 0 with curl/wget)

siedenburg2 (commented):
Changed the simple server swap to a solution where the server tries to download from any url in the array and if the file on the server isn't the right mimetype ("application/octet-stream" or "application/x-gzip") or if the server is offline etc, it's counted to an error counter and if every url in the array fails the installation will be aborted.

liaralabs:
Awesome stuff, thank you!

Oudinunit:
Ok ka Thanks you ka my family

เมื่อ อา. 14 ม.ค. 2567 เวลา 01:49 liaralabs ***@***.***>
เขียนว่า:

PR #1144, opened 2024-03-23
rTorrent: Increase address space to 4GB with 32GB of RAM

stickz:
This commit allows the rtorrent chunk manager to temporarily allocate up of 4GB of memory, on systems with more than 32GB of available memory. This is useful for increasing download throughput. It will not allocate the memory unless it's available.


stickz:
Not working properly.

PR #1234, opened 2025-06-01
rTorrent: Patch two 0.15.3 crashes

stickz:
This commit patches two rTorrent `0.15.3` software crashes. These changes have been linked and merged upstream.


liaralabs:
Since a new release has been pushed, we should support that instead, rather than patching .3

stickz:
> Since a new release has been pushed, we should support that instead, rather than patching .3

There are two crashes on `0.15.4` we need to patch. These were found 2 hours after release. 

I will submit a pull request for `0.15.4`. But it's critical we patch and support `0.15.3`. It's been tested with a thousand users on docker and I got it stable.

